PR #22, opened 2018-10-31
Lagom module

ygree:
Mainly copy-paste of @chbatey work from https://github.com/lightbend/lagom-couchbase/pull/1
Plus added missing bare minimum part to run Java HelloWorld write-side example.


patriknw:
Perhaps retain attribution to @chbatey by using `git commit --author`?

ygree:
> Perhaps retain attribution to @chbatey by using `git commit --author`?

Good idea! Fixed!

PR #31, opened 2018-11-01
Add notes about the primary index

ygree:


ygree:
We have decided to create and remove the bucket for the tests automatically, so no need for this index

Issue #7, opened 2018-10-30
Configurable bucket


johanandren:
Right now we have hardcoded "akka" in a few places, make sure that comes from config.

Issue #6, opened 2018-10-30
Load settings from config


johanandren:
Initially we'll keep the settings hardcoded, to move forward quickly, but when the set of configuration options have stabilized we should make them loadable from the system config.

We should make sure to avoid the somewhat messy config situation we have in Cassandra, so make sure to look into that and what we have learned from that.

PR #38, opened 2018-11-02
Making settings load from file

johanandren:
Refs #6 and #7


core/src/main/scala/akka/persistence/couchbase/settings.scala
```diff
@@ -0,0 +1,56 @@
...
+}
+
+final class CouchbaseReadJournalSettings private (
+  val sessionSettings: CouchbaseSessionSettings,
+  val bucket:          String)
+
+object CouchbaseReadJournalSettings {
+
+  def apply(config: Config): CouchbaseReadJournalSettings = {
+    // FIXME uses the same config as CouchbaseJournalSettings for now
```
patriknw: whatever you do, don't separate them as we have done in akka-persistence-cassandra

couchbase-client/src/main/scala/akka/stream/alpakka/couchbase/settings.scala
```diff
@@ -0,0 +1,138 @@
...
+    val password = config.getString("password")
+    val nodes = config.getStringList("nodes").asScala.toList
+    new CouchbaseSessionSettings(username, password, nodes, None)
+  }
+
+  def apply(username: String, password: String): CouchbaseSessionSettings = {
+    new CouchbaseSessionSettings(username, password, Nil, None)
+  }
+
+  def create(username: String, password: String): CouchbaseSessionSettings = {
```
patriknw: `create` with `Config` also

couchbase-client/src/main/scala/akka/stream/alpakka/couchbase/settings.scala
```diff
@@ -0,0 +1,138 @@
...
+
+  val default = apply()
+
+  // FIXME default params in public API is problematic for bincomp
+  def apply(
+    parallelism: Int                           = 1,
+    replicateTo: ReplicateTo                   = ReplicateTo.ONE,
+    persistTo:   PersistTo                     = PersistTo.NONE,
+    timeout:     Long                          = 2L,
+    timeUnit:    java.util.concurrent.TimeUnit = java.util.concurrent.TimeUnit.SECONDS): CouchbaseWriteSettings =
```
patriknw: why not FintiteDuration/Duration?

couchbase-client/src/main/scala/akka/stream/alpakka/couchbase/settings.scala
```diff
@@ -0,0 +1,138 @@
...
+import com.typesafe.config.Config
+
+import scala.collection.JavaConverters._
+import scala.collection.immutable
+
+object CouchbaseWriteSettings {
+
+  val default = apply()
+
+  // FIXME default params in public API is problematic for bincomp
```
patriknw: yes, only include mandatory fields here, and others can be amended using `withX`

couchbase-client/src/main/scala/akka/stream/alpakka/couchbase/settings.scala
```diff
@@ -0,0 +1,138 @@
...
+  def apply(
+    parallelism: Int                           = 1,
+    replicateTo: ReplicateTo                   = ReplicateTo.ONE,
+    persistTo:   PersistTo                     = PersistTo.NONE,
```
patriknw: what are these replicateTo and persistTo used for, and are these default safe to use in production?

couchbase-client/src/main/scala/akka/stream/alpakka/couchbase/settings.scala
johanandren: It's just following the internal type from the client, copy-pasta from the Alpakka PR. I think it may need to keep the time unit, which java duration does not do, or maybe that doesn't matter...

couchbase-client/src/main/scala/akka/stream/alpakka/couchbase/settings.scala
johanandren: You only use these explicitly with the `.insert(thing, writeSettings)`, I haven't looked into the consequences of them yet, this is also from the Alpakka PR

core/src/main/scala/akka/persistence/couchbase/settings.scala
johanandren: Yeah, I made a note in #6 about avoiding the config file mess in Cassandra. It's not 100% clear for me what a good solution would be though. Do we have ideas about how we would wish to deal with it for Cassandra written down?

couchbase-client/src/main/scala/akka/stream/alpakka/couchbase/settings.scala
patriknw: I'm not sure how it's used, but if the underlying api requires unit you can always convert to millis

couchbase-client/src/main/scala/akka/stream/alpakka/couchbase/settings.scala
patriknw: ok, I don't think write consistencies should have defaults, especially not weak ones like these. better to make it an explicit decision.

core/src/main/scala/akka/persistence/couchbase/settings.scala
patriknw: not decided yet, but some notes in https://github.com/akka/akka-persistence-cassandra/issues/81

I think we only have to support configuration of both sides together.

One can always create another plugin instance with other config for special needs.

core/src/main/scala/akka/persistence/couchbase/CouchbaseJournal.scala
```diff
@@ -95,25 +94,23 @@ Need at least this GI:
...
 
   private val log = Logging(this)
   private val serialization: Serialization = SerializationExtension(context.system)
+  private implicit val materializer = ActorMaterializer()
 
-  private implicit val materializer = ActorMaterializer() // FIXME keeping our own rather than reusing another one?
+  private val settings: CouchbaseJournalSettings = {
+    // shared config is one level above the journal specific
+    val commonPath = configPath.replaceAll("""\.write$""", "")
+    val sharedConfig = context.system.settings.config.getConfig(commonPath)
```
patriknw: good

johanandren:
Ready for final review. (I think we should deal with the default write settings or not in a separate PR, created #55 to track that for now)

core/src/main/resources/reference.conf
```diff
@@ -1,32 +1,28 @@
-akka.persistence.couchbase {
-
+couchbase-journal {
```
chbatey: the reason i didn't use this initially was that https://github.com/Product-Foundry/akka-persistence-couchbase but I don't think that matters

core/src/main/resources/reference.conf
```diff
@@ -1,32 +1,28 @@
...
+  # connection settings shared by write, read and snapshot
   connection {
-    hosts = "localhost"
+    nodes = ["localhost"]
+    # username = "admin"
+    # password = "admin1"
```
chbatey: super secret

core/src/main/scala/akka/persistence/couchbase/CouchbaseJournal.scala
```diff
@@ -95,25 +94,23 @@ Need at least this GI:
...
   private implicit val ec: ExecutionContext = context.dispatcher
 
   private val log = Logging(this)
   private val serialization: Serialization = SerializationExtension(context.system)
+  private implicit val materializer = ActorMaterializer()
 
-  private implicit val materializer = ActorMaterializer() // FIXME keeping our own rather than reusing another one?
+  private val settings: CouchbaseJournalSettings = {
+    // shared config is one level above the journal specific
+    val commonPath = configPath.replaceAll("""\.write$""", "")
```
chbatey: Thinking about this some more what if users define a custom config location for the journal not called write. I guess that just won't be supported

core/src/main/resources/reference.conf
johanandren: Wouldn't expect that you'd use both in the same project, right?

core/src/main/scala/akka/persistence/couchbase/CouchbaseJournal.scala
johanandren: They'll always have to make the custom path one up, since we want to read the `connection` section in the parent block.

PR #40, opened 2018-11-04
Fix a bug: snapshot deserialization issue.

ygree:
CouchbasePersistentEntityActorSpec.saveSnapshot was failing.

### DIFF
--- a/core/src/main/scala/akka/persistence/couchbase/CouchbaseSnapshotStore.scala
+++ b/core/src/main/scala/akka/persistence/couchbase/CouchbaseSnapshotStore.scala
@@ -69,7 +69,8 @@ class CouchbaseSnapshotStore(cfg: Config) extends SnapshotStore {
     )
 
     result.map {
-      case Some(value) =>
+      case Some(snapshot) =>
+        val value = snapshot.getObject("akka")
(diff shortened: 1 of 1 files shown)


core/src/main/scala/akka/persistence/couchbase/CouchbaseSnapshotStore.scala
```diff
@@ -67,16 +67,15 @@ class CouchbaseSnapshotStore(cfg: Config) extends SnapshotStore {
...
-    }
+    result.map(_.map { snapshot =>
+      val value = snapshot.getObject("akka")
+      SelectedSnapshot(
+        SnapshotMetadata(
+          persistenceId,
+          value.getLong(Fields.SequenceNr),
+          value.getLong(Fields.Timestamp)),
+        Serialized.fromJsonObject(serialization, value))
+    })
```
johanandren: I think the pattern matching was somewhat cleaner, to be honest. The actual bugfix is reaching into the "akka" field?

ygree: Yes. Fixed.

Issue #53, opened 2018-11-08
Replays must be consistent


johanandren:
When replaying a persistent actor we must always see the last event written by a previous instance.

PR #59, opened 2018-11-08
Configure replication and persist requirements

johanandren:
Refs #53

Just started out with some settings for now


core/src/main/resources/reference.conf
```diff
@@ -10,6 +10,36 @@ couchbase-journal {
...
+    parallelism = 1
+
+    # How many nodes a write is replicated to before considering the write completed.
+    #
+    # Valid values:
+    # * none - No replication needed to consider a write done
+    # * one - Write replicated to at least one replica
+    # * two - Write replicated to at least two replicas
+    # * three - Write replicated to at least three replicas
+    replicate-to = none
```
patriknw: is it possible to use safe defaults here?

I guess `one` would work also for a single node local setup, and it's at least better than `none`.

(we don't want to get the mongodb stamp)

core/src/main/resources/reference.conf
```diff
@@ -10,6 +10,36 @@ couchbase-journal {
...
+    # How many nodes a write has reached disk on before it is considered complete.
+    #
+    # Valid values:
+    # * none - Write reaching memory is enough
+    # * master - Ensure the modification is stored on disk on the master node
+    # * one - Ensure the modification is stored on disk on at least one nodes
+    # * two - Ensure the modification is stored on disk on at least two nodes
+    # * three- Ensure the modification is stored on disk on at least three nodes
+    # * four - Ensure the modification is stored on disk on the active node and all three replicas
+    persist-to = none
```
patriknw: same here

core/src/main/scala/akka/persistence/couchbase/CouchbaseJournal.scala
```diff
@@ -110,10 +121,41 @@ class CouchbaseJournal(config: Config, configPath: String) extends AsyncWriteJou
...
+        .eq(x("$1"))
+        .and(x(Fields.SequenceFrom).gte(x("$2")))
+    )
+    .orderBy(desc(Fields.SequenceFrom))
+    .limit(1)
+
+  // FIXME, needs an orderby make sure index works with it
+  // select * from akka where akka.persistenceId = "pid1" and sequence_from >= start and sequenceFrom <= end
+  // can't be a val because we need to limit and that mutates the statement :/
+  private def replayStatement =
```
patriknw: `val`

core/src/main/scala/akka/persistence/couchbase/CouchbaseJournal.scala
```diff
@@ -148,75 +190,80 @@ class CouchbaseJournal(config: Config, configPath: String) extends AsyncWriteJou
...
   }
 
+  // FIXME also backoff - currently thunders along and keeps failing
+  // FIXME potentially not viable with persistTo higher than none
+  private def nextOrdering(numberOfIndexes: Int): Future[Long] =
+    couchbase
+      .counter(OrderingCounter, numberOfIndexes, 0, settings.writeSettings)
+      .recoverWith {
+        case _: DurabilityException =>
+          // concurrent update to counter, retry
```
patriknw: so it's the global counter that is causing it?

core/src/main/scala/akka/persistence/couchbase/CouchbaseJournal.scala
chbatey: i don't think they are re-useable

core/src/main/scala/akka/persistence/couchbase/CouchbaseJournal.scala
chbatey: ah that sounds plausable, but it isn't like is is a CAS so that is a bit crap

core/src/main/resources/reference.conf
johanandren: One means one more, so that would not work with a one-node cluster out of the box (might be fine)

core/src/main/resources/reference.conf
johanandren: One does not currently work, that's why it is left to none for now. Requiring a write causes the contention on the ordering counter to fail tests, even with a retry in place. We need to figure that out.

core/src/main/scala/akka/persistence/couchbase/CouchbaseJournal.scala
johanandren: See the comment just above

core/src/main/scala/akka/persistence/couchbase/CouchbaseJournal.scala
johanandren: Yes, exactly, the global counter + a persist requirement rather than just in memory update makes it so that we get stuck in a state where all writes retry and fail. Let's discuss this monday.

core/src/main/resources/reference.conf
johanandren: I wonder if we should perhaps not have defaults at all, since with an actual cluster "safe" could be `replicate-to` > none or `persist-to` > none or some combination of both.

If we have defaults I'd prefer that the integration tests actually run with them, and that doesn't pass with anything else than none/none right now. 

`persist-to` > 0 seems insanely slow (times out individual inserts), `replicate-to` > 0 is much much faster, but I'm not sure we can run more than 1 node on travis. 

How do we deal with this for Cassandra, do we always run a n-node Cassandra cluster for the tests?

johanandren: Or, when I think of it, we just skip the perf-spec on travis, because the other ones pass.

johanandren: persist-to one as default and used in tests now.

core/src/main/scala/akka/persistence/couchbase/CouchbaseJournal.scala
patriknw: my bad

core/src/main/resources/reference.conf
johanandren: Ok, good then I think `persist-to = one` fulfill  that same purpose here.

johanandren: Or even better since it is somewhat safe also with an actual multi-node cluster.

patriknw: ok, good. We can document production considerations.

johanandren:
Ready for final review, last commit after Patriks ok is just docker and readme

PR #64, opened 2018-11-10
[WIP] Implemented backpressure test

ygree:


couchbase-client/src/test/scala/akka/stream/alpakka/couchbase/BackpressureTestSpec.scala
```diff
@@ -0,0 +1,122 @@
...
+      val stmt = select("*").from(bucket.name())
+      val query = N1qlQuery.simple(
+        stmt, queryParams
+      )
+
+      val result = session
+        .streamedQuery(query)
+        .map(d => d.getObject(bucketName).getInt("id"))
+        .fold(0L)((acc, i) => acc + i)
+        .runWith(Sink.last)
```
johanandren: We wouldn't need to read all of them to verify back pressure works, so we should be able to also do a `throttle(1.second)` and then `take(a-few)` and that should show if it works or not.

Issue #54, opened 2018-11-08
Query backpressure


johanandren:
We need to investigate if this somehow works out of the box (I don't think so, because rxJava), cover with some test coverage and if it doesn't work, see if we can solve it through pagination or something like that (https://blog.couchbase.com/optimizing-database-pagination-using-couchbase-n1ql/)

ygree:
I’ve been trying to prove that Couchbase Java SDK doesn’t support back-pressure by trying to break it.
I thought there was no back pressure because I saw tcp packets kept going when it was still consuming the result set to the very end even when the stream was paused. It happens because it has enough internal buffers to accommodate lots of data and in my case there was relatively small amount of data requested (100k records). In fact when I run 12 concurrent queries returning 1M documents each they all finish successfully. It takes some time but it returns all the records correctly with no failures when the timeouts set large enough. I think since it uses tcp connection it effectively uses tcp back-pressure.

Couchbase Java SDK uses Netty under the hood. I was trying to minimize all internal buffers by setting all the Netty allocators settings to the minimum:

```System.setProperty("com.couchbase.client.deps.io.netty.allocator.pageSize", "4096") // minimum 4096
System.setProperty("com.couchbase.client.deps.io.netty.allocator.numHeapArenas", "0")
System.setProperty("com.couchbase.client.deps.io.netty.allocator.numDirectArenas", "0")
System.setProperty("com.couchbase.client.deps.io.netty.allocator.maxOrder", "0")
System.setProperty("com.couchbase.client.deps.io.netty.allocator.maxCachedBufferCapacity", "0")
System.setProperty("com.couchbase.client.deps.io.netty.allocator.cacheTrimInterval", "0")
System.setProperty("com.couchbase.client.deps.io.netty.allocator.useCacheForAllThreads", "false")
System.setProperty("com.couchbase.client.deps.io.netty.allocator.tinyCacheSize", "0")
System.setProperty("com.couchbase.client.deps.io.netty.allocator.smallCacheSize", "0")
System.setProperty("com.couchbase.client.deps.io.netty.allocator.normalCacheSize", "0")```


But despite all these it still works fine.

What’s more, when the stream fails it nicely cancels and it stops requesting data, e.g.:

```],
"status": "closed",
"metrics": {"elapsedTime": "1.2361925s","executionTime": "1.23615s","resultCount": 14646,"resultSize": 906416}
}```


All these leads me to the conclusion that there is in fact back-pressure in place.

[...]

PR #67, opened 2018-11-12
Fail the stream on error

ygree:
Addresses this issue: https://github.com/akka/akka-persistence-couchbase/issues/54#issuecomment-437559744


couchbase-client/src/test/scala/akka/stream/alpakka/couchbase/RequestTimeoutTestSpec.scala
```diff
@@ -0,0 +1,132 @@
...
+
+  "Couchbase Java SDK" should {
+
+    "drop index to improve insertion time" in {
+      bucket.bucketManager().dropN1qlPrimaryIndex(true)
+    }
+
+    "clean up db" in {
+      bucket.bucketManager().flush()
+    }
```
johanandren: I'd put these in before instead of as test steps, we're not testing these method of the sdk

couchbase-client/src/test/scala/akka/stream/alpakka/couchbase/RequestTimeoutTestSpec.scala
```diff
@@ -0,0 +1,132 @@
...
+        .runWith(Sink.ignore)
+
+      assert(result.futureValue === Done)
+    }
+
+    "create index" in {
+      bucket.bucketManager().createN1qlPrimaryIndex(true, false)
+    }
+
+    "fail the stream with request timeout" in {
```
johanandren: I think you can put this combined test case in the CouchbaseSessionSpec instead since it is about testing the session.

couchbase-client/src/main/scala/akka/stream/alpakka/couchbase/impl/RxUtilities.scala
```diff
@@ -29,8 +26,17 @@ private[couchbase] object RxUtilities {
...
-      t.rows().map(unfoldDocument)
+    def call(t: AsyncN1qlQueryResult): Observable[JsonObject] = {
+      val data: Observable[JsonObject] = t.rows().map(unfoldDocument)
+      val errors = t
+        .errors()
+        .flatMap(new Func1[JsonObject, Observable[JsonObject]] {
+          override def call(
+              err: JsonObject
+          ): Observable[JsonObject] = Observable.error(CouchbaseResponseException(err))
+        })
```
johanandren: This function we can create a single instance and cache/reuse just like `unfoldJsonObjects` rather than create a new one on each invocation

couchbase-client/src/main/scala/akka/stream/alpakka/couchbase/CouchbaseResponseException.scala
```diff
@@ -0,0 +1,19 @@
+/*
+ * Copyright (C) 2018 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.stream.alpakka.couchbase
+import com.couchbase.client.java.document.json.JsonObject
+
+class CouchbaseResponseException(val msg: String, val code: Int) extends RuntimeException {
```
johanandren: "If you liked it, make it final" ;)

Also, runtime exception already has a message field so pass that to the super constructor

couchbase-client/src/main/scala/akka/stream/alpakka/couchbase/CouchbaseResponseException.scala
```diff
@@ -0,0 +1,19 @@
...
+
+  override def toString = s"CouchbaseResponseException($msg, $code)"
+}
+
+object CouchbaseResponseException {
```
johanandren: `/** INTERNAL API */ @InternalApi private[akka]` since this is for parsing output of couchbase rather than something a user would interact with.

couchbase-client/src/main/scala/akka/stream/alpakka/couchbase/CouchbaseResponseException.scala
```diff
@@ -0,0 +1,19 @@
...
+  def apply(json: JsonObject): CouchbaseResponseException =
+    new CouchbaseResponseException(
+      msg = if (json.containsKey("msg")) json.getString("msg") else "",
+      code = if (json.containsKey("code")) json.getInt("code") else 0
```
johanandren: Make code an option if it may be missing?

couchbase-client/src/test/scala/akka/stream/alpakka/couchbase/RequestTimeoutTestSpec.scala
```diff
@@ -0,0 +1,132 @@
...
+import com.couchbase.client.java.query.{N1qlParams, N1qlQuery}
+import org.scalatest.concurrent.{Eventually, ScalaFutures}
+import org.scalatest.{BeforeAndAfterAll, Matchers, WordSpec}
+
+import scala.concurrent.duration._
+
+class RequestTimeoutTestSpec extends WordSpec with Matchers with ScalaFutures with BeforeAndAfterAll with Eventually {
+
+  // FIXME use a automagic docker-couchbase
+  // currently requires a running couchbase with an admin account named 'admin' with password 'admin1'
```
johanandren: This is the same as the other tests though? Ticketify what it would mean to do this automagically?

couchbase-client/src/main/scala/akka/stream/alpakka/couchbase/CouchbaseResponseException.scala
ygree: It should always be there but I'll make it Option as there is no guarantee

couchbase-client/src/test/scala/akka/stream/alpakka/couchbase/RequestTimeoutTestSpec.scala
ygree: Copy-pasted. It's not relevant anymore since we've got Travis setup.

johanandren: Please remove then

couchbase-client/src/test/scala/akka/stream/alpakka/couchbase/RequestTimeoutTestSpec.scala
ygree: I made it a separate test because it won't work with the other tests because this test requires specific CouchbaseEnvironment setup that is needed to reproduce timeout error.

PR #72, opened 2018-11-14
Support for async serializers 

### LINKED ISSUE #8: Support async serializers
We need support for async serializers (for the commercial LB GDPR support)

johanandren:
Fixes #8


johanandren:
I think the future got through to the user side somehow, do'h

johanandren:
Now it should work, but would like to do some refactoring on top, feels like a bit of a mess with how the serialization/deserialization logic is spread out currently.

johanandren:
Still not 100% happy but that will have to do for now.

core/src/main/scala/akka/persistence/couchbase/CouchbaseSchema.scala
```diff
@@ -0,0 +1,213 @@
...
+    val sequenceTo: Long = value.getLong(Fields.SequenceTo)
+    val localTo: Long = if (sequenceTo > toSequenceNr) toSequenceNr else sequenceTo
+    val events: JsonArray = value.getArray(Fields.Messages)
+    val writerUuid = value.getString(Fields.WriterUuid)
+    require(sequenceTo - from + 1 == events.size())
+    val nrReply = localTo - from
+    Future.sequence((0 to nrReply.asInstanceOf[Int]).map { i =>
+      extract(persistenceId, writerUuid, events.getObject(i), serialization)
+    }.toVector)
+  }
```
johanandren: This might not look like it belongs here, but is is used from the read journal as well

core/src/main/scala/akka/persistence/couchbase/CouchbaseJournal.scala
```diff
@@ -159,63 +106,74 @@ class CouchbaseJournal(config: Config, configPath: String) extends AsyncWriteJou
...
+      .recover {
+        // lift the failure to not fail the whole sequence
+        case NonFatal(ex) => Failure(ex)
+      }(ExecutionContexts.sameThreadExecutionContext)
+
+  private def atomicWriteToJsonDoc(write: AtomicWrite): Future[JsonDocument] = {
+    // collect all tags for the events and put in the surrounding
+    // object for simple indexing on tags
+    // FIXME we can't do this anymore, we're juggling futures below
+    var allTags = Set.empty[String]
```
johanandren: This I need to fix before I'm done with this PR

johanandren: Fixed

PR #73, opened 2018-11-14
Serialize to couchbase native json

### LINKED ISSUE #14: Investigate if native JSON storage is possible and useful
Would it be possible to serialize into nested JSON in the documents and would it be useful in any way? Note that we will anyway have to support non-nested-json for arbitrary serializers.

johanandren:
Fixes #14


johanandren:
Ready for final review

core/src/main/scala/akka/persistence/couchbase/CouchbaseSchema.scala
```diff
@@ -108,12 +110,22 @@ private[akka] final object CouchbaseSchema {
...
       .put(Fields.SerializerManifest, msg.manifest)
       .put(Fields.SerializerId, msg.identifier)
-      .put(Fields.Payload, Base64.getEncoder.encodeToString(msg.payload))
+
+    msg.nativePayload match {
+      case OptionVal.None =>
+        json.put(Fields.Payload, Base64.getEncoder.encodeToString(msg.payload))
+
+      case OptionVal.Some(jsonPayload) =>
+        json.put(Fields.Payload, jsonPayload)
```
patriknw: could it be confusing to use the same field for both binary and json payload? let's say someone wants to setup an index into the json payload (but ignore documents with binary payload)?

README.md
```diff
@@ -89,6 +89,18 @@ If you will be using the query side with event-for-tags the following will also
 CREATE INDEX `tags` ON `akka`((all (`all_tags`)),`ordering`);
 ```
 
+## Serialization
+
+The couchbase journal uses the Akka serialization infrastructure to serialize events and snapshots into bytes
+which is then Base64 encoded and put inside the documents as strings. This is convenient as it allows re-use of exisiting
```
chbatey: which are then

core/src/main/scala/akka/persistence/couchbase/CouchbaseSchema.scala
```diff
@@ -180,33 +201,48 @@ private[akka] object SerializedMessage {
...
         Serialization.withTransportInformation(system.asInstanceOf[ExtendedActorSystem]) { () =>
-          async.toBinaryAsync(event)
+          async
+            .toBinaryAsync(event)
+            .map(bytes => SerializedMessage(serId, serManifest, bytes, OptionVal.None))(
+              ExecutionContexts.sameThreadExecutionContext
+            )
         }
+
+      case jsonSerializer: JsonSerializer =>
```
chbatey: AsyncJsonSerializer? Guess we don't need that

core/src/main/scala/akka/persistence/couchbase/JsonSerializer.scala
```diff
@@ -0,0 +1,33 @@
...
+ * Copyright (C) 2018 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.persistence.couchbase
+import akka.serialization.SerializerWithStringManifest
+import com.couchbase.client.java.document.json.JsonObject
+import com.couchbase.client.java.transcoder.JacksonTransformers
+
+/**
+ * Special serializer base class to use Couchbase native json for serializing the events and snapshots.
```
chbatey: :+1:

core/src/main/scala/akka/persistence/couchbase/CouchbaseSchema.scala
johanandren: Nah, the only purpose of it is the GDPR stuff, and that won't work anyways.

PR #75, opened 2018-11-15
[spike] Auto create index async

ygree:


core/src/main/scala/akka/persistence/couchbase/CouchbaseJournal.scala
```diff
@@ -120,7 +119,9 @@ class CouchbaseJournal(config: Config, configPath: String) extends AsyncWriteJou
     CouchbaseJournalSettings(sharedConfig)
   }
 
-  private val couchbase = CouchbaseSession(settings.sessionSettings, settings.bucket)
+  println("CouchbaseJournal.indexAutoCreate = " + settings.indexAutoCreate)
+  private val couchbase =
+    CouchbaseSessionFactory(context.system, settings.sessionSettings, settings.bucket, settings.indexAutoCreate)
```
johanandren: I think we should do like this (after some discussions with the team and thinking about it):

Make `CouchbaseSession.async` the default `apply` when the user does not already provide a bucket. If the user wants it blocking they can use with `Await`.

That will make this a `Future[CouchbaseSession]`, this will allow us to make sure it is not completed until index is created:

```
val couchbase = {
  val session = CouchbaseSession(stuff)
  if (settings.autoCreateIndex) session.flatMap(_.createIndex(...)) 
  else session
}
```
Put the index details in the new `CouchbaseSchema` though, since we have that now.

then when we interact with it in the journal we can shortcut it, to use a `flatMap` if it isn't completed or just do the operation if the session is completed.

Something like: 
```
  def withCouchbase[A](f: CouchbaseSession => Future[A]): Future[A] = {
    couchbase.value match {
      case Some(Success(c)) => f(c)
      case _ => couchbase.flatMap(f)
    }
  }
```

We also need to deal with failure of that future, so that it leads to the journal failing.

ygree:
Address all the comments in the other PR: https://github.com/akka/akka-persistence-couchbase/pull/81

PR #79, opened 2018-11-15
Release setup

### LINKED ISSUE #52: Set up for releasing
I think we can ask M2 for help to get some nice tag-to-release infra in place.

johanandren:
Fixes #52

Based on the alpakka-kafka-connector release setip.


build.sbt
```diff
@@ -1,27 +1,36 @@
...
+  scmInfo := Some(ScmInfo(url("https://github.com/akka/akka-persistence-couchbase"), "https://github.com/akka/akka-persistence-couchbase.git")),
   startYear := Some(2018),
+  developers += Developer("contributors",
+    "Contributors",
+    "https://gitter.im/akka/dev",
+    url("https://github.com/akka/akka-persistence-couchbase/graphs/contributors")),
   licenses := Seq(("Apache-2.0", url("http://www.apache.org/licenses/LICENSE-2.0"))),
-  crossScalaVersions := Seq("2.11.12", "2.12.7"),
+  crossScalaVersions := Seq("2.12.7", "2.11.12"),
   scalaVersion := crossScalaVersions.value.last,
```
patriknw: shouldn't 2.12 be default?

johanandren: We've stumbled on a few issues using 2.12-only future features by default, so I thought it was better to catch those early

.travis.yml
```diff
@@ -56,6 +56,10 @@ stages:
   - name: whitesource
     if: repo = akka/akka-persistence-couchbase AND ( ( branch = master AND type = push ) OR tag =~ ^v )
 
+  # runs on main repo master commits and version-tagged commits
+  - name: publish
+    if: repo = akka/reactive-kafka AND ( ( branch = master AND type = push ) OR tag =~ ^v )
```
johanandren: Does this mean we will publish each main commit to bintray @2m ?

Do we need some cleanup job as well?

RELEASING.md
```diff
@@ -0,0 +1,16 @@
+# Releasing
+		
+ From a direct clone (rather than a fork). You will need permission in sonatype to push to akka repositories.
```
chbatey: is this true given all you do is create a tag?

build.sbt
chbatey: I prefer 2.12 as the default with a travis building 2.11 but I don't mind

.travis.yml
chbatey: should we just do tags?

RELEASING.md
johanandren: The first sentence not sure, but second you need to be able to push that tag to the repo.

2m:
The actual job definition is missing:
```
jobs:
  include:
    ...
    - stage: publish
      script: sbt +publish
```

FYI: the `stages` section is for defining conditions when a particular stage is ran. And the `jobs` section for defining the jobs themselves.

RELEASING.md
2m: You can create tags from GitHub UI which will kick start the release.

Permissions to publish to the sonatype group id `com.lightbend.akka` will be required to push the sync button in bintray that syncs artifacts from bintray to sonatype repos.

```suggestion
 From a direct clone (rather than a fork). You will need permission in sonatype to push to `com.lightbend.akka` groupd ID.
```

build.sbt
```diff
@@ -1,27 +1,40 @@
 import sbt.Keys.{name, publishArtifact}
 
-crossScalaVersions := Seq("2.12.6", "2.11.12")
+resolvers in ThisBuild ++= Seq(Resolver.bintrayRepo("manub", "maven"))
```
2m: This is not needed here. We use this resolver to pull in `"net.manub" %% "scalatest-embedded-kafka" % "2.0.0"` dependency.

.travis.yml
johanandren: Thanks for clarifying, I'll leave it like this for now then

build.sbt
johanandren: Ah, wasn't sure, but the maven one is still needed?

PR #81, opened 2018-11-15
Avoid blocking code and Index autocreate

ygree:
https://github.com/akka/akka-persistence-couchbase/pull/81


core/src/main/scala/akka/persistence/couchbase/Couchbase.scala
```diff
@@ -0,0 +1,73 @@
...
+import akka.stream.alpakka.couchbase.scaladsl.CouchbaseSession
+import akka.stream.scaladsl.Source
+import com.couchbase.client.java.Bucket
+import org.slf4j.LoggerFactory
+
+import scala.concurrent.{ExecutionContext, Future}
+import scala.util.Success
+
+object Couchbase {
+  private val log = LoggerFactory.getLogger(this.getClass)
```
johanandren: This one could likely be blocking, better use the Akka logging infrastructure

core/src/main/scala/akka/persistence/couchbase/Couchbase.scala
```diff
@@ -0,0 +1,73 @@
...
+      case Some(Success(c)) => f(c)
+      case _ =>
+        Source.fromFuture(underlying).flatMapConcat(f) //TODO: is there a way to preserve Mat type of f here?
+    }
+
+  def close(): Unit =
+    //leaving it closing behind for now
+    underlying.foreach(_.close())
+
+}
```
johanandren: I don't think these names aren't conveying what they do very well, I think `withCouchbase` or `performCouchbaseOperation` would be better. 

Maybe it would be smoother with this as a mixin trait, requiring a `def asyncCouchbaseSession` and providing the methods directly on this?

core/src/main/scala/akka/persistence/couchbase/Couchbase.scala
```diff
@@ -0,0 +1,73 @@
...
+    underlying.value match {
+      case Some(Success(c)) => f(c)
+      case _ => underlying.flatMap(f)
+    }
+
+  def mapToSource[Out](f: CouchbaseSession => Source[Out, NotUsed]): Source[Out, NotUsed] =
+    underlying.value match {
+      case Some(Success(c)) => f(c)
+      case _ =>
+        Source.fromFuture(underlying).flatMapConcat(f) //TODO: is there a way to preserve Mat type of f here?
```
johanandren: Yep, should definitely be possible.

core/src/main/scala/akka/persistence/couchbase/CouchbaseReadJournal.scala
```diff
@@ -24,13 +23,6 @@ object CouchbaseReadJournal {
   final val Identifier = "couchbase-journal.read"
 }
 
-/*
-Required indexes:
-
-CREATE INDEX `pi2` ON `akka`((self.`persistenceId`),(self.`sequence_from`))
-
-
- */
```
johanandren: I think a note that it requires the index still makes sense.

core/src/main/scala/akka/persistence/couchbase/CouchbaseReadJournal.scala
```diff
@@ -51,10 +43,11 @@ class CouchbaseReadJournal(eas: ExtendedActorSystem, config: Config, configPath:
 
     CouchbaseReadJournalSettings(sharedConfig)
   }
-  private val session = CouchbaseSession(settings.sessionSettings, settings.bucket)
+
+  private val couchbase = Couchbase(settings.sessionSettings, settings.bucket, settings.indexAutoCreate)
 
   system.registerOnTermination {
-    session.close()
+    couchbase.close()
```
johanandren: I wonder if we should really do this at all,  we have a poststop hook already and the journal is an actor, so it will always stop before system terminates.

core/src/main/scala/akka/persistence/couchbase/CouchbaseSchema.scala
```diff
@@ -157,6 +160,13 @@ private[akka] final object CouchbaseSchema {
       )
     }
 
+  /**
+   * Creates `pi2` index
+   *
+   * @return true if the index is already existed, otherwise false
+   */
+  def createPi2Index(session: CouchbaseSession): Future[Boolean] =
+    session.createIndex("pi2", true, Fields.PersistenceId, Fields.SequenceFrom)
```
johanandren: There should be one more index here, in a separate method, the tags one. And then only create that when the read journal starts (no need for it if that isn't used, I think).

core/src/main/scala/akka/persistence/couchbase/Couchbase.scala
```diff
@@ -0,0 +1,73 @@
...
+import akka.stream.alpakka.couchbase.CouchbaseSessionSettings
+import akka.stream.alpakka.couchbase.scaladsl.CouchbaseSession
+import akka.stream.scaladsl.Source
+import com.couchbase.client.java.Bucket
+import org.slf4j.LoggerFactory
+
+import scala.concurrent.{ExecutionContext, Future}
+import scala.util.Success
+
+object Couchbase {
```
johanandren: AsyncCouchbaseSession/AsyncCouchbaseWrapper/AsyncSessionWrapper something like that rather, this doesn't say much about what it does?

core/src/main/scala/akka/persistence/couchbase/CouchbaseJournal.scala
```diff
@@ -115,13 +113,14 @@ class CouchbaseJournal(config: Config, configPath: String) extends AsyncWriteJou
...
-        case NonFatal(ex) => Failure(ex)
-      }(ExecutionContexts.sameThreadExecutionContext)
+    couchbase.mapToFuture(
+      _.insert(jsonDoc, settings.writeSettings)
+        .map(json => ExtraSuccessFulUnit)(ExecutionContexts.sameThreadExecutionContext)
+        .recover {
+          // lift the failure to not fail the whole sequence
+          case NonFatal(ex) => Failure(ex)
+        }(ExecutionContexts.sameThreadExecutionContext)
+    )
```
johanandren: Here in the use site I also think something like:

```
withCouchbase { couchbase => 
   couchbase.doStuff()
}
``` 

would be more clear

couchbase-client/src/main/scala/akka/stream/alpakka/couchbase/scaladsl/CouchbaseSession.scala
```diff
@@ -23,25 +24,32 @@ object CouchbaseSession {
...
-    val cluster: Cluster =
-      settings.environment match {
-        case Some(environment) => CouchbaseCluster.create(environment, settings.nodes: _*)
-        case None => CouchbaseCluster.create(settings.nodes: _*)
-      }
-    cluster.authenticate(settings.username, settings.password)
-    val bucket = cluster.openBucket(bucketName)
-    new CouchbaseSessionImpl(bucket, Some(cluster))
+    //wrap CouchbaseAsyncCluster.create up in the Future because it's blocking
+    val asyncCluster: Future[AsyncCluster] =
```
chbatey: The async client blocks on creation? Imagine this is initialised during startup for normal use cases so that is okay

core/src/main/scala/akka/persistence/couchbase/Couchbase.scala
chbatey: I like the with* sounds good

core/src/main/scala/akka/persistence/couchbase/CouchbaseSchema.scala
chbatey: called pi2 as it was my second go at it? :)

couchbase-client/src/main/scala/akka/stream/alpakka/couchbase/scaladsl/CouchbaseSession.scala
johanandren: I think we may need to isolate it on a separate dispatcher if that is the case, so that we do not cause a deadlock/starvation on startup if you run the system with few threads.

core/src/main/scala/akka/persistence/couchbase/CouchbaseSchema.scala
johanandren: I see, let's give it a better name also then. :)

PR #82, opened 2018-11-16
Additional work on top of the async-client and indexes

johanandren:
Mostly applying my own comments on #81


johanandren:
Biggest changes are transforming `Couchbase` class and companion into a single `AsyncCouchbaseSession` trait, moving the index responsibility to the respective plugins that needs the indexes.

core/src/main/resources/reference.conf
```diff
@@ -40,8 +40,8 @@ couchbase-journal {
     # Fail a read from couchbase if it takes longer than this
     read-timeout = 15 s
 
-    # Parameter indicating whether the `pi2` index should be auto created if it does not exist
-    # DO NOT USE THIS IN PRODUCTION
+    # Auto create the required indexes for the write journal if they are missing when the
```
chbatey: I wonder if we shouldn't do this and just rely on the global index that comes with a default couchbase install?

chbatey: then we can avoid some of the init

core/src/main/scala/akka/persistence/couchbase/CouchbaseReadJournal.scala
```diff
@@ -300,6 +325,6 @@ class CouchbaseReadJournal(eas: ExtendedActorSystem, config: Config, configPath:
 
 
    */
-  override def persistenceIds(): Source[String, NotUsed] =
-    ???
+  // override def persistenceIds(): Source[String, NotUsed] =
```
chbatey: just delete?

core/src/main/resources/reference.conf
chbatey: And maybe have a class they can run separately that creates the indexes that we can use in tests?

johanandren: Yeah, maybe, it _is_ really useful also for just playing around with the journal with minimal setup fuzz though.

core/src/main/scala/akka/persistence/couchbase/CouchbaseReadJournal.scala
johanandren: We'll implement it next week or so, I just thought it was unfair to claim we support it by implementing the trait with a stub.

Issue #76, opened 2018-11-15
Hook in Lagom json serialization into native Couchbase serialization


johanandren:
Since Lagom uses Jackson/Play json it would be great if we could hook that right into the journal and get that json into the couchbase AST out of the box with Lagom

ygree:
I've tried to convert to/from Couchbase `JsonObject`s but Couchbase Java SDK has it's own version of Jackson `com.couchbase.client.deps.com.fasterxml.jackson.databind.module` that's binary incompatible with Lagom's version and packages are renamed.

Another idea is to get rid of Couchbase’s `JsonObject` default representation and define our own Json payload representation (with ability to inject underlying user’s serialization result as String). Couchbase Java SDK allows to define a custom document representation and mark it as Json document for N1QL support. It can be done by implementing a custom `transcoder` (see example https://gist.github.com/daschl/a42fa84ac7b01e627c3a40299d9f93cd#file-bytejsontranscoder-java-L29). It will require changes in the core module as it relies on Couchbase’s `JsonObject`

PR #85, opened 2018-11-19
[WIP] Jackson native serializer spike

ygree:


johanandren:
Kind of refs to #76

lagom-persistence-couchbase/javadsl/src/main/scala/com/lightbend/lagom/javadsl/persistence/couchbase/JacksonCouchbaseSerializer.scala
```diff
@@ -0,0 +1,103 @@
...
+    val clazz = system.dynamicAccess.getClassFor[AnyRef](className) match {
+      case Success(clazz) ⇒ clazz
+      case Failure(e) ⇒
+        throw new NotSerializableException(
+          s"Cannot find manifest class [$className] for serializer [${getClass.getName}]."
+        )
+    }
+
+    val jsonStr = json.toString //TODO avoid String cross conversion
+    objectMapper.readValue(jsonStr, clazz)
```
johanandren: Yeah, avoiding this is the tricky part :)

lagom-persistence-couchbase/javadsl/src/main/scala/com/lightbend/lagom/javadsl/persistence/couchbase/JacksonCouchbaseSerializer.scala
```diff
@@ -0,0 +1,103 @@
...
+  override def manifest(obj: AnyRef): String = {
+    val className = obj.getClass.getName
+    migrations.get(className) match {
+      case Some(transformer) => className + "#" + transformer.currentVersion
+      case None => className
+    }
+  }
+
+  override def toJson(o: AnyRef): JsonObject = {
+    val jsonStr = objectMapper.writeValueAsString(o)
```
patriknw: I think that as long as compression is not used the bytes returned from Lagom's JacksonJsonSerializer can be converted to a string with `new String(bytes, "UTF-8")`

patriknw: I think all this can be simplified by delegating to Lagom's `JacksonJsonSerializer` instead of reimplementing it. The bytes should be possible to convert to/from strings.

lagom-persistence-couchbase/javadsl/src/main/scala/com/lightbend/lagom/javadsl/persistence/couchbase/JacksonCouchbaseSerializer.scala
ygree: I've tried to convert to/from Couchbase JsonObjects but it has it's own version of Jackson `com.couchbase.client.deps.com.fasterxml.jackson.databind.module` that's binary incompatible with Lagom's version and packages are renamed.

lagom-persistence-couchbase/javadsl/src/main/scala/com/lightbend/lagom/javadsl/persistence/couchbase/JacksonCouchbaseSerializer.scala
patriknw: Looks like this feature might be more complicated than we thought. Perhaps not worth doing for Lagom, and instead pick it up again when we add JacksonJsonSerializer to Akka?

johanandren: Let's see if everything else is works out smooth enough that we have time to take another look at this before the full release.

lagom-persistence-couchbase/javadsl/src/main/scala/com/lightbend/lagom/javadsl/persistence/couchbase/JacksonCouchbaseSerializer.scala
ygree: There is one more idea is to get rid of Couchbase’s JsonObject default representation and define our own Json payload representation (possibly with ability to inject underlying user’s serialization result as String). Couchbase Java SDK allows to define a custom document representation and mark it as Json document for N1QL support. It can be done by implementing a custom transcoder (see example https://gist.github.com/daschl/a42fa84ac7b01e627c3a40299d9f93cd#file-bytejsontranscoder-java-L29). It will require changes in the core module as it relies on Couchbase’s JsonObject.

johanandren:
Let's close this for now as we are not actively working on it

PR #89, opened 2018-11-22
UUID based ordering

### LINKED ISSUE #66: Order for event by tag that scales
The previous impl used a counter which will, when configured with persistTo > 0 or replicateTo > 0 fail concurrent requests. One idea is timestamps which will work as long as there isn't clock skew.

johanandren:
Fixes #66

A string representation for the time based event UUID that sorts the same as actual UUIDs would (and the comparator does) and contains all the information needed to recreate the UUID instances in memory.


core/src/main/scala/akka/persistence/couchbase/UUIDGenerator.scala
```diff
@@ -0,0 +1,294 @@
...
+  }
+
+}
+
+/**
+ * Scala implementation of UUIDComparator in
+ * https://github.com/cowtowncoder/java-uuid-generator
+ * Apache License 2.0.
+ */
+class TimeBasedUUIDComparator extends Comparator[UUID] {
```
johanandren: We'll see if we actually need this for anything else than the spec checking ordering

core/src/main/scala/akka/persistence/couchbase/UUIDGenerator.scala
```diff
@@ -0,0 +1,294 @@
...
+}
+
+/**
+ * INTERNAL API
+ */
+@InternalApi
+object UUIDGenerator {
+
+  private val MinLSB = 0x0L
+  private val MaxLsb = Long.MaxValue
```
johanandren: Not sure we'll need this one

johanandren: Ah, we will need it for read-behind later.

core/src/main/scala/akka/persistence/couchbase/UUIDGenerator.scala
```diff
@@ -0,0 +1,294 @@
...
+
+      // If the clock went back in time (currentTimeMillis is not monotonic) just pick next and ids will hopefully catch up
+      // I think this should increase the clock part (according to spec), following Cassandra impl here
+      if (now.toMs < lastMs) {
+        if (lastTimeStamp.compareAndSet(last.nanoTimestamp, candidate.nanoTimestamp)) candidate
+        else currentTimestamp() // lost the cas, try again
+      } else {
+        // We trying to pick more than one ids in the same currentTimeMillis resolution interval
+        // Generating more than 10k uuids in one millisecond is not likely with the Couchbase journal
+        // so if that happens we just fail
```
patriknw: "just fail"? where is that?

core/src/main/scala/akka/persistence/couchbase/CouchbaseReadJournal.scala
```diff
@@ -207,20 +219,33 @@ class CouchbaseReadJournal(eas: ExtendedActorSystem, config: Config, configPath:
...
         }
       }
       .mapConcat(identity)
+      .statefulMapConcat[EventEnvelope] { () =>
+        // deduplicate
+        var lastUUID = UUIDGenerator.MinUUID
+
+        { (elem) =>
+          elem.offset match {
+            case TimeBasedUUID(uuid) if lastUUID == uuid => Nil
```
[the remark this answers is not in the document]
johanandren: Nope, this was a left-over from the first iteration when I thought we couldn't get away from it.

core/src/main/scala/akka/persistence/couchbase/UUIDGenerator.scala
```diff
@@ -0,0 +1,288 @@
...
+
+  val GMT = ZoneId.of("GMT")
+
+  // UUID v1 timestamp must be in 100-nanoseconds interval since Oct 15 1582 00:00:00.000
+  // this is the offset of that from the unix timestamp starting point Jan 01 1970 00:00:00.000
+  private val StartEpochMillis = -12219292800000L
+
+  val MinVal = UUIDTimestamp(Long.MinValue)
+
+  def now(): UUIDTimestamp = apply(System.currentTimeMillis())
```
patriknw: That apply is from the `case class`. right? Then `currentTimeMillis` is used as the nanoTimestamp parameter, but the other `apply extracts StartEpochMillis and some other stuff

johanandren: Indeed, good catch. Should be `fromUnixTimestamp`

johanandren: Didn't notice since it is only used from currentTimestamp which increments and cas-spins

core/src/main/scala/akka/persistence/couchbase/UUIDGenerator.scala
```diff
@@ -0,0 +1,288 @@
...
+    .appendValue(ChronoField.SECOND_OF_MINUTE, 2)
+    .appendLiteral('.')
+    // include nano of second instead of standard millis to not lose precision
+    .appendValue(ChronoField.NANO_OF_SECOND, 9)
+    .toFormatter()
+    .withZone(UUIDTimestamp.GMT)
+
+  def extractMac(): Long = {
+    // use first interface we find
+    def firstAwailableMac(enumeration: java.util.Enumeration[NetworkInterface]): Array[Byte] =
```
chbatey: available

core/src/main/scala/akka/persistence/couchbase/UUIDGenerator.scala
```diff
@@ -0,0 +1,288 @@
...
+        else firstAwailableMac(enumeration)
+      } else {
+        Array.emptyByteArray
+      }
+    val interfaces = NetworkInterface.getNetworkInterfaces()
+    val mac = firstAwailableMac(interfaces)
+    if (mac.isEmpty)
+      // spec says to generate pseudo random if there is no mac but we don't need to care about that case
+      // because if we can connect to a db there is an interface
+      throw new RuntimeException("Could not find any network interfaces to base UUIDs on")
```
chbatey: Fair enough!

chbatey: Interesting that the C* driver doesn't use the NetworkAddress with comments implying it isn't possible

core/src/main/scala/akka/persistence/couchbase/UUIDGenerator.scala
```diff
@@ -0,0 +1,288 @@
...
+  private val lastTimeStamp = new AtomicLong(0L)
+
+  /*
+   * Note that currently we use {@link System#currentTimeMillis} for a base time in
+   * milliseconds, which is not monotonic and might have less than 1 ms resolution, on top of that
+   * the precision we need is 100-nanosecond intervals we can only
+   * generate at most 10K UUID if the resolution of currentTimeMillis is one millisecond safely.
+   */
+  @tailrec
+  def currentTimestamp(): UUIDTimestamp = {
```
chbatey: what's stopping two instances on the same node generating the same UUIDs? Is that what the `Random.nextInt()` is for?

core/src/main/scala/akka/persistence/couchbase/UUIDGenerator.scala
johanandren: Yeah, I wonder if it predates that API (Java 8?) perhaps?

core/src/main/scala/akka/persistence/couchbase/UUIDGenerator.scala
johanandren: Yeah, the spec says if "it is too expensive to share a persistent counter across processes, use random for 'clock'" or something along those lines. Not watertight but good enough? I started out with putting the pid in there, but that doesn't help if you run multiple nodes in the same JVM.

core/src/main/scala/akka/persistence/couchbase/CouchbaseReadJournal.scala
```diff
@@ -148,80 +153,74 @@ class CouchbaseReadJournal(eas: ExtendedActorSystem, config: Config, configPath:
...
+        CouchbaseSchema.deserializeEvents(value, Long.MaxValue, serialization, CouchbaseSchema.extractEvent).map {
           deserialized =>
             deserialized.map(
               tpr =>
                 EventEnvelope(Offset
-                                .sequence(tpr.pr.sequenceNr), // FIXME, should this be +1, check inclusivity of offsets
-                              tpr.pr.persistenceId,
-                              tpr.pr.sequenceNr,
-                              tpr.pr.payload)
+                                .sequence(tpr.sequenceNr), // FIXME, should this be +1, check inclusivity of offsets
```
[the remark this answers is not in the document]
johanandren: Yeah, there's a separate issue for that #90

core/src/main/scala/akka/persistence/couchbase/CouchbaseReadJournal.scala
```diff
@@ -148,80 +153,74 @@ class CouchbaseReadJournal(eas: ExtendedActorSystem, config: Config, configPath:
...
           )
           .mapMaterializedValue(_ => NotUsed)
 
-      eventsByTagSource(
-        sourceOfRows,
-        tag
-      )
+      @volatile var lastUUID = UUIDGenerator.MinUUID
+
+      taggedRows.mapAsync(1) { row: AsyncN1qlQueryRow =>
```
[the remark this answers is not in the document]
johanandren: Created #94 to track that

core/src/main/scala/akka/persistence/couchbase/settings.scala
```diff
@@ -51,26 +61,59 @@ object CouchbaseJournalSettings {
...
-    new CouchbaseReadJournalSettings(sessionSettings, bucket)
+    val pagesize = config.getInt("read.page-size")
+
+    val eventByTagConfig = config.getConfig("read.events-by-tag")
+    val eventByTagSettings = new EventByTagSettings(
+      eventByTagConfig.getDuration("eventual-consistency-delay").toMillis.millis,
+      eventByTagConfig.getBoolean("verbose-debug-logging")
+    )
+
+    val scanConsistency = config.getString("read.scan-consistency").toLowerCase match {
```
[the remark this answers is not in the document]
johanandren: I think I'll remove it again based on Patriks feedback in the other PR

core/src/main/resources/reference.conf
```diff
@@ -43,7 +43,25 @@ couchbase-journal {
...
+    page-size = 100
+
+
+    # The consistency requirement to apply to the queries - valid values are:
+    # "not-bounded", "request-plus", "statement-plus" - for more details see
+    # https://docs.couchbase.com/java-sdk/2.7/scan-consistency-examples.html
+    scan-consistency = request-plus
+
     events-by-tag {
+      # For eventsByTag queries how long to delay the query for. For event writes that come from different nodes
```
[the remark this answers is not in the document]
johanandren: I think I copied it from the C* journal ;)

core/src/main/scala/akka/persistence/couchbase/internal/N1qlQueryStage.scala
```diff
@@ -52,13 +57,15 @@ class N1qlQueryStage[S](live: Boolean,
     var rowsInCurrentQuery = 0
     // TODO use a mutable buffer e.g. ArrayDeque
     var buffer = Vector.empty[AsyncN1qlQueryRow]
+    var lastRow = null
```
ygree: Can't see any usages of `lastRow`

PR #104, opened 2018-11-30
Persistence ids

### LINKED ISSUE #90: persistenceIds query
Needs to be implemented

Missing:
* offset start exclusivity
* offset ends inside of batches
* could potentially be better implemented with unesting and skipping the seq_from/seq_to fields

johanandren:
On top of the eventsByPersistenceId query PR so review that one first.

Fixes #90 

Live query works but have to think a bit about paging for the non-live one: offset based paging and random order of persistence ids means queries could potentially miss persistence ids.

### DIFF
--- a/core/src/test/scala/akka/persistence/couchbase/CouchbaseReadJournalSpec.scala
+++ b/core/src/test/scala/akka/persistence/couchbase/CouchbaseReadJournalSpec.scala
@@ -68,6 +68,34 @@ class CouchbaseReadJournalSpec
     }
   }
 
+  "live persistenceIds" must {
+    "show new persistence ids" in {
+      val senderProbe = TestProbe()
+      implicit val sender = senderProbe.ref
+
+      val queryProbe: TestSubscriber.Probe[String] =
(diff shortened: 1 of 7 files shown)


johanandren:
Rebased on master

johanandren:
Let's leave the paging for `currentPersistenceIds` for now, I created a separate issue #108 to track it

core/src/test/scala/akka/persistence/couchbase/CouchbaseReadJournalSpec.scala
```diff
@@ -68,6 +68,34 @@ class CouchbaseReadJournalSpec
...
+
+      awaitAssert({
+        queryProbe.expectNext("p3")
+      }, 5.seconds)
+
+      val pa4 = system.actorOf(TestActor.props("p4"))
+      pa4 ! "p4-evt-1"
+      senderProbe.expectMsg("p4-evt-1-done")
+
+      // we shouldn't see p3 again
```
chbatey: expectNoMessage as they could come out of order?

PR #105, opened 2018-11-30
Session registry in an Akka extension

### LINKED ISSUE #77: Re-use the same session across plugins
Currently we create a session each in the `CouchbaseJournal`, `CouchbaseSnapshotStore` and `CouchbaseReadJournal`, if the settings are the same we should share the same instance. This can be done through an extension - could even live in the connector.

No need to prioritize this until we have at least two more complete journal/stores.

johanandren:
Fixes #77


core/src/main/scala/akka/persistence/couchbase/CouchbaseSessionExtension.scala
```diff
@@ -0,0 +1,58 @@
...
+package akka.persistence.couchbase
+import java.util.concurrent.atomic.AtomicReference
+
+import akka.actor.{ExtendedActorSystem, Extension, ExtensionId}
+import akka.stream.alpakka.couchbase.CouchbaseSessionSettings
+import akka.stream.alpakka.couchbase.scaladsl.CouchbaseSession
+
+import scala.concurrent.{Future, Promise}
+
+object CouchbaseSessionExtension extends ExtensionId[CouchbaseSessionExtension] {
```
patriknw: normal way of implementing extension is
```
object CouchbaseSessionExtension extends ExtensionId[CouchbaseSessionExtension] with ExtensionIdProvider {
```

and then override `get` and `lookup`.

Better name might be `CouchbaseSessionRegistry`

core/src/main/scala/akka/persistence/couchbase/CouchbaseSessionExtension.scala
```diff
@@ -0,0 +1,58 @@
...
+
+  private def startSession(key: SessionKey): Future[CouchbaseSession] = {
+    val promise = Promise[CouchbaseSession]()
+    val oldSessions = sessions.get()
+    val newSessions = oldSessions.updated(key, promise.future)
+    if (sessions.compareAndSet(oldSessions, newSessions)) {
+      // we won cas, initialize session
+      system.log.info("Starting Couchbase session for nodes [{}]", key.settings.nodes.mkString("\"", "\", \"", "\""))
+      promise.completeWith(CouchbaseSession(key.settings, key.bucketName))
+      // FIXME should we hook in a system termination that closes the session as well?
```
patriknw: yes

core/src/main/scala/akka/persistence/couchbase/CouchbaseSessionExtension.scala
johanandren: Dang, I never quite remember that API...

patriknw: impossible to remember, copy is my friend

couchbase-client/src/main/scala/akka/stream/alpakka/couchbase/CouchbaseSessionRegistry.scala
```diff
@@ -0,0 +1,93 @@
...
+  private val log = Logging(system, classOf[CouchbaseSessionRegistry])
+
+  private val sessions = new AtomicReference(Map.empty[SessionKey, Future[CouchbaseSession]])
+
+  /**
+   * Scala API: Get an existing session or start a new one with the given settings and bucket name,
+   * makes it possible to share one session across plugins.
+   *
+   * Note that the session must not be stopped manually, it is shut down when the actor system is shutdown,
+   * if you need a more fine grained life cycle control, create the CouchbaseSession manually instead.
```
chbatey: Do it your self fool!

couchbase-client/src/main/scala/akka/stream/alpakka/couchbase/CouchbaseSessionRegistry.scala
```diff
@@ -0,0 +1,93 @@
...
+      system.registerOnTermination {
+        future.foreach { session =>
+          session.close()
+          log.info("Shutting down couchbase session for nodes [{}]", nodesAsString)
+        }(system.dispatcher)
+      }
+      future
+    } else {
+      // we lost cas, someone else is initializing session
+      sessions.get()(key)
```
chbatey: may have been for a different key, call recursively?

couchbase-client/src/main/scala/akka/stream/alpakka/couchbase/CouchbaseSessionRegistry.scala
```diff
@@ -0,0 +1,93 @@
...
+   * if you need a more fine grained life cycle control, create the CouchbaseSession manually instead.
+   */
+  def getSessionFor(settings: CouchbaseSessionSettings, bucketName: String): CompletionStage[JCouchbaseSession] =
+    sessionFor(settings, bucketName).map(_.asJava)(system.dispatcher).toJava
+
+  private def startSession(key: SessionKey): Future[CouchbaseSession] = {
+    val promise = Promise[CouchbaseSession]()
+    val oldSessions = sessions.get()
+    val newSessions = oldSessions.updated(key, promise.future)
+    if (sessions.compareAndSet(oldSessions, newSessions)) {
```
chbatey: a ConcurrentMap and computeIfAbsent may be simpler?

johanandren: That doesn't guarantee that only one session is created while this does.

couchbase-client/src/main/scala/akka/stream/alpakka/couchbase/CouchbaseSessionRegistry.scala
johanandren: Ah, good point

couchbase-client/src/main/scala/akka/stream/alpakka/couchbase/CouchbaseSessionRegistry.scala
patriknw: computeIfAbsent could do that

this map is typically populated at startup and then it might be less overhead with AtomicRef for subsequent access

johanandren: From the Javadoc of `computeIfAbsent`:

> The default implementation may retry these steps when multiple
> threads attempt updates including potentially calling the mapping
> function multiple times.

Which means concurrent calls to `computeIfAbsent` it could create multiple client sessions just to throw all but one away (and leak those since close wouldn't be called).

patriknw: Where?

My ConcurrentHashMap says:
```
     * The entire
     * method invocation is performed atomically, so the function is
     * applied at most once per key.  Some attempted update operations
     * on this map by other threads may be blocked while computation
     * is in progress, so the computation should be short and simple,
     * and must not attempt to update any other mappings of this map.
```

johanandren: Ah, `ConcurrentMap` vs `ConcurrentHashMap`

couchbase-client/src/main/scala/akka/stream/alpakka/couchbase/CouchbaseSessionRegistry.scala
```diff
@@ -0,0 +1,93 @@
...
+ * rather than creating a new Couchbase session per use site.
+ */
+object CouchbaseSessionRegistry extends ExtensionId[CouchbaseSessionRegistry] with ExtensionIdProvider {
+  def createExtension(system: ExtendedActorSystem): CouchbaseSessionRegistry =
+    new CouchbaseSessionRegistry(system)
+
+  /**
+   * Java API:
+   */
+  override def get(system: _root_.akka.actor.ActorSystem): CouchbaseSessionRegistry =
```
patriknw: proper improt

couchbase-client/src/main/scala/akka/stream/alpakka/couchbase/CouchbaseSessionRegistry.scala
johanandren: Should I switch? (I see that it has two synchronized blocks in there on Java 8 at least)

patriknw: I think I prefer the AtomicReference for this since it's updated infrequently and we anyway want Future values in there. I think ConcurrentHashMap has slower access and more blocking.

Issue #106, opened 2018-12-03
Make CouchbaseReadJournal public


patriknw:
CouchbaseReadJournal is marked as Internal API. The read journals are given much freedom to implement additional queries and I think we should follow that here also, i.e. make it public. Put it in scaladsl and javadsl packages. Let CouchbaseReadJournalProvider return the CouchbaseReadJournal.

If for nothing else the exact semantics of the queries must be documented and that is best done as Scaladoc/Javadoc.

Follow same pattern as in Cassandra plugin.

If you don't by these arguments, it should be moved to the `internal` package.

PR #112, opened 2018-12-03
Make CouchbaseReadJournal public and implement its JavaDsl version

ygree:
Refs #106, #102


core/src/main/scala/akka/persistence/couchbase/javadsl/CouchbaseReadJournal.scala
```diff
@@ -0,0 +1,213 @@
...
+   * The stream is completed with failure if there is a failure in executing the query in the
+   * backend journal.
+   */
+  override def eventsByTag(tag: String, offset: Offset): Source[EventEnvelope, NotUsed] =
+    scaladslReadJournal.eventsByTag(tag, offset).asJava
+
+  /**
+   * Same type of query as `eventsByTag` but the event stream
+   * is completed immediately when it reaches the end of the "result set". Events that are
+   * stored after the query is completed are not included in the event stream.
```
johanandren: This sentence is a bit weird, for any query new results won't be in it after it is completed. For events by persistence id there is a test that says that events written after the query is triggered/executed are not included (but it fails now and then) - there is no test coverage for that for tags though. Added #113 to track that.

core/src/main/scala/akka/persistence/couchbase/javadsl/CouchbaseReadJournal.scala
```diff
@@ -0,0 +1,213 @@
...
+   * To acquire an offset from a long unix timestamp to use with this query, you can use [[timeBasedUUIDFrom]].
+   *
+   */
+  override def currentEventsByTag(tag: String, offset: Offset): Source[EventEnvelope, NotUsed] =
+    scaladslReadJournal.currentEventsByTag(tag, offset).asJava
+
+  /**
+   * Same type of query as `persistenceIds` but the event stream
+   * is completed immediately when it reaches the end of the "result set". Events that are
+   * stored after the query is completed are not included in the event stream.
```
johanandren: Same phrasing issue here, not completed, but triggered, although in this case it is currently true because there is no paging here yet so it is a single query.

core/src/test/scala/akka/persistence/couchbase/CouchbaseReadJournalSpec.scala
```diff
@@ -161,6 +162,53 @@ class CouchbaseReadJournalSpec
...
+      probe1.expectNextPF { case e @ EventEnvelope(_, "a", 2L, "a green apple") => e }
+      val offs = probe1.expectNextPF { case e @ EventEnvelope(_, "a", 4L, "a green banana") => e }.offset
+      probe1.cancel()
+
+      val greenSrc2 = queries.currentEventsByTag(tag = "green", offs)
+      val probe2 = greenSrc2.runWith(TestSink.probe[Any])
+      probe2.request(10)
+      probe2.expectNextPF { case e @ EventEnvelope(_, "b", 2L, "a green leaf") => e }
+      probe2.cancel()
+    }
```
johanandren: This already covered by `find existing events with an offset into a batch` and `find existing events with an offset into multiple batches`, but let's keep the tests anyways, can't hurt.

If we want to test the UUID factory that should also be a simple unit test rather than an integration test.

core/src/test/scala/akka/persistence/couchbase/javadsl/CouchbaseReadJournalSpec.scala
```diff
@@ -0,0 +1,112 @@
...
+      val src = javaQueries.currentEventsByTag("a", Offset.noOffset)
+      src.asScala
+        .map(_.persistenceId)
+        .runWith(TestSink.probe[Any])
+        .request(10)
+        .expectNext("a")
+        .expectComplete()
+    }
+  }
+}
```
johanandren: We should test the Java APIs with Java code

core/src/test/scala/akka/persistence/couchbase/scaladsl/CouchbaseReadJournalSpec.scala
```diff
@@ -0,0 +1,111 @@
...
+    with WordSpecLike
+    with CouchbaseBucketSetup
+    with Matchers
+    with Eventually {
+
+  implicit val mat = ActorMaterializer()(system)
+
+  lazy val queries = PersistenceQuery(system).readJournalFor[CouchbaseReadJournal](CouchbaseReadJournal.Identifier)
+
+  "Couchbase Read Journal Java API" must {
```
johanandren: Duplicated tests? Shouldn't this contain the current `CouchbaseReadJournalSpec` ?

core/src/main/scala/akka/persistence/couchbase/javadsl/CouchbaseReadJournal.scala
```diff
@@ -0,0 +1,213 @@
...
+   * known `persistenceId`s is provided by `currentPersistenceIds`.
+   *
+   * Note the query is inefficient, especially for large numbers of `persistenceId`s, because
+   * of limitation of current internal implementation providing no information supporting
+   * ordering/offset queries. The query uses Couchbase's `select distinct` capabilities.
+   * More importantly the live query has to repeatedly execute the query each `refresh-interval`,
+   * because order is not defined and new `persistenceId`s may appear anywhere in the query results.
+   */
+  override def persistenceIds(): Source[String, NotUsed] = scaladslReadJournal.persistenceIds().asJava
+}
```
johanandren: Good docs in general!

core/src/main/scala/akka/persistence/couchbase/CouchbaseReadJournalProvider.scala
```diff
@@ -5,13 +5,16 @@
 package akka.persistence.couchbase
 
 import akka.actor.ExtendedActorSystem
-import akka.persistence.query.scaladsl.ReadJournal
-import akka.persistence.query.{javadsl, ReadJournalProvider}
+import akka.persistence.couchbase.javadsl.{CouchbaseReadJournal => JavaDslCouchbaseReadJournal}
+import akka.persistence.couchbase.scaladsl.{CouchbaseReadJournal => ScalaDslCouchbaseReadJournal}
```
patriknw: perhaps remove these imports and use `javadsl.CouchbaseReadJournal` and `scaladsl.CouchbaseReadJournal` as the types below

core/src/main/scala/akka/persistence/couchbase/scaladsl/CouchbaseReadJournal.scala
```diff
@@ -0,0 +1,370 @@
...
+    with EventsByTagQuery
+    with CurrentEventsByTagQuery
+    with CurrentPersistenceIdsQuery
+    with PersistenceIdsQuery
+    with CouchbaseSchema.Queries {
+
+  private implicit val system = eas
+  private val log = Logging(system, configPath)
+
+  protected implicit val executionContext = system.dispatcher
```
patriknw: should probably be the plugin dispatcher
```
system.dispatchers.lookup(queryPluginConfig.pluginDispatcher)
```

core/src/main/scala/akka/persistence/couchbase/javadsl/CouchbaseReadJournal.scala
ygree: Good catch

core/src/test/scala/akka/persistence/couchbase/scaladsl/CouchbaseReadJournalSpec.scala
ygree: Right. Removed and moved scaladsl specific test into scaladsl package

core/src/main/scala/akka/persistence/couchbase/scaladsl/CouchbaseReadJournal.scala
ygree: Good catch! Thanks, Patrik!

core/src/test/scala/akka/persistence/couchbase/CouchbaseReadJournalSpec.scala
ygree: Good suggestion. Add UUIDsSpec in the separate PR

PR #115, opened 2018-12-04
Aligning the client with Alpakka

### LINKED ISSUE #37: Complete the client/connector
The current client/connector is the minimal set of features the journal needs and does not cover all the Source/Flow/Sink:s that the Alpakka connector does and does not yet have any Java API. If we want to merge this connector I think we must cover those.

johanandren:
Since there already is an Alpakka Couchbase connector that we intend to replace with the Couchbase client part part we want to make sure we have feature parity with the existing one (not necessarily source compatibility though).

Fixes #37


couchbase-client/src/main/scala/akka/stream/alpakka/couchbase/impl/CouchbaseSessionImpl.scala
```diff
@@ -39,10 +40,15 @@ final private[couchbase] class CouchbaseSessionImpl(asyncBucket: AsyncBucket, cl
 
   override def underlying: AsyncBucket = asyncBucket
 
-  def insert(document: JsonDocument): Future[JsonDocument] =
+  def insert(document: JsonDocument): Future[JsonDocument] = insertDoc(document)
+
+  def insertDoc[T <: Document[_]](document: T): Future[T] =
     singleObservableToFuture(asyncBucket.insert(document), document)
 
   def insert(document: JsonDocument, writeSettings: CouchbaseWriteSettings): Future[JsonDocument] =
```
patriknw: is `JsonDocument` a `Document`? then why is both signatures needed?

johanandren: It's mentioned in the public API Scaladoc: when using the generic in streams, the type inference fails in Scala and Java, so I added a separate method for the "normal" case when you work with JSON docs, and then for the special other document types you will have to be more explicit.

PR #119, opened 2018-12-05
UUIDsSpec

ygree:


core/src/test/scala/akka/persistence/couchbase/UUIDsSpec.scala
```diff
@@ -0,0 +1,38 @@
...
+import org.scalatest.{Matchers, WordSpec}
+import org.scalatest.concurrent.ScalaFutures
+import org.scalatest.prop.Checkers
+
+class UUIDsSpec extends WordSpec with Matchers with ScalaFutures with Checkers {
+
+  val timestamp = Gen.choose(
+    UUIDTimestamp.MinVal.toUnixTimestamp,
+    UUIDTimestamp.MaxVal.toUnixTimestamp / 10 // TODO: there is some overflow happening for bigger values
+  ).filter(_ != 0)
```
johanandren: The UUID timepart is a long expressing hundreds of nanos since Oct 15 1582 00:00:00.000, which means that values higher than 5236-03-31T21:21:00.684006975 cannot be expressed, while unix timestamp max is 292278994-08-17 07:12:55 meaning that passing arbitrary unix timestamp could fail, same goes for minimum, that doesn't really matter in our use case though.

Starting from the max of uuid timestamp shouldn't be a problem, so if that's what you are seeing there is a bug somewhere.

core/src/test/scala/akka/persistence/couchbase/UUIDsSpec.scala
```diff
@@ -0,0 +1,38 @@
...
+      }
+    )
+    "preserve timestamp" in check(
+      forAll(timestamp) { ts =>
+        val uuid = UUIDs.timeBasedUUIDFrom(ts).asInstanceOf[TimeBasedUUID]
+        UUIDs.timestampFrom(uuid) === ts
+      }
+    )
+  }
+}
```
johanandren: I wouldn't have felt it is worth pulling in Scalacheck for such a simple thing but ok.

The bounds checking rather belongs in the existing UUIDSpec which already has some tests for the UUIDTimestamp - where the actual conversion logic lives.

ygree: I'll remove the scalacheck dependency

core/src/test/scala/akka/persistence/couchbase/UUIDsSpec.scala
ygree: I've tried roundtrip conversion for the max value and it seems to be broken. See "roundtrip max possible UUID back into itself"

core/src/test/scala/akka/persistence/couchbase/internal/UUIDSpec.scala
```diff
@@ -24,6 +24,16 @@ class UUIDSpec extends WordSpec with Matchers with ScalaFutures {
...
+    "roundtrip a max unix timestamp back into itself" in {
+      val unixTimestamp = UUIDTimestamp.MaxVal.toUnixTimestamp
+      val result = UUIDTimestamp.fromUnixTimestamp(unixTimestamp).toUnixTimestamp
+      result should ===(unixTimestamp)
+    }
+    "roundtrip a min unix timestamp back into itself" in {
+      val unixTimestamp = UUIDTimestamp.MinVal.toUnixTimestamp
+      val result = UUIDTimestamp.fromUnixTimestamp(unixTimestamp).toUnixTimestamp
+      result should ===(unixTimestamp)
+    }
```
johanandren: 👍

johanandren:
Closing this as I have continued in a new PR

PR #122, opened 2018-12-06
Optional warnings about missing indexes

### LINKED ISSUE #18: At least warn about missing indices
It is silent about missing indices. It behaves like there are no events when the indices are missing.

johanandren:
Fixes #18


core/src/main/scala/akka/persistence/couchbase/CouchbaseJournal.scala
```diff
@@ -88,6 +88,23 @@ class CouchbaseJournal(config: Config, configPath: String)
...
   }
 
+  if (settings.warnAboutMissingIndexes) {
+    for {
+      session <- asyncSession
+      indexes <- session.listIndexes().runWith(Sink.seq)
+    } {
+      val indexNames = indexes.map(_.name()).toSet
+      Set("persistence-ids", "sequence-nrs").foreach { requiredIndex =>
+        println(requiredIndex + " " + indexNames(requiredIndex))
```
ygree: Should be `log.info` instead?

johanandren: Ups, left over, shouldn't be here at all :)

PR #131, opened 2018-12-07
Index check also for snapshot plugin

johanandren:
Missed this one in the previous index-check PR


core/src/main/scala/akka/persistence/couchbase/CouchbaseSnapshotStore.scala
```diff
@@ -45,6 +47,21 @@ final class CouchbaseSnapshotStore(cfg: Config, configPath: String) extends Snap
...
+  if (settings.warnAboutMissingIndexes) {
+    implicit val materializer = ActorMaterializer()
+    for {
+      session <- asyncSession
+      indexes <- session.listIndexes().runWith(Sink.seq)
+    } {
+      materializer.shutdown() // not needed after used
+      val indexNames = indexes.map(_.name()).toSet
+      if (!indexNames("snapshots"))
+        log.error(
```
patriknw: should it only be logged? or should it be done as part of the initialization of the `Future[CouchbaseSession]` and thereby failing that Future?

core/src/main/scala/akka/persistence/couchbase/CouchbaseSnapshotStore.scala
```diff
@@ -45,6 +47,21 @@ final class CouchbaseSnapshotStore(cfg: Config, configPath: String) extends Snap
     log.error(ex, "Failed to connect to couchbase")
     context.stop(self)
   }
+  if (settings.warnAboutMissingIndexes) {
+    implicit val materializer = ActorMaterializer()
+    for {
+      session <- asyncSession
+      indexes <- session.listIndexes().runWith(Sink.seq)
+    } {
+      materializer.shutdown() // not needed after used
```
patriknw: this is not done if any of the above futures failed, leaking a materializer

core/src/main/scala/akka/persistence/couchbase/CouchbaseSnapshotStore.scala
johanandren: I don't think it is worth the work to do it that way because of how invasive it is compared to the value of this feature, even if it would be somewhat more "correct".

core/src/main/scala/akka/persistence/couchbase/CouchbaseSnapshotStore.scala
johanandren: If they fail it will be because the session couldn't start which will stop the actor, which is the ActorRefFactory the materializer was started with so it should be ok.

johanandren: Added #140 to track that for the query side

Issue #23, opened 2018-10-31
Lagom docs


ygree:
See https://github.com/lightbend/lagom-couchbase/pull/1#issuecomment-419989312

PR #137, opened 2018-12-07
Ported the Lagom Java docs to Scala

johanandren:
* Made the sample classes actually compile
 * Moved around the Lagom pages a bit in the docs structure (no sub category)

Refs #23


johanandren:
I'll ask someone from the Lagom team to read through the docs next week and see if there are any issues (there's also one outstanding example that is just a place holder)

docs/src/test/scala/docs/home/persistence/CouchbaseReadSideProcessor.scala
```diff
@@ -0,0 +1,102 @@
...
+    override def aggregateTags: Set[AggregateEventTag[HelloEvent]] =
+      // TODO return the tag for the events
+      ???
+  }
+  //#initial
+}
+
+object CouchbaseReadSideProcessorTwo {
+
+  class HelloEventProcessor(readSide: ReadSide with CouchbaseReadSide)(implicit executionContext: ExecutionContext)
```
ygree: `ReadSide` is the existing registry to register a read-side processor, whereas `CouchbaseReadSide` is a specific read-side builder injected into a concrete read-side processor implementation. Here it should be `readSide: CouchbaseReadSide`

johanandren: Ups this is a mistake, the problem I had was in the GreetingServiceImpl  when I tried to mimic the Lagom docs here where I need both register from one of the types and then the other type to pass here. Should the scaladsl CouchbaseReadSide extend ReadSide and that is the issue perhaps?

docs/src/test/scala/docs/home/persistence/CouchbaseReadSideQuery.scala
```diff
@@ -0,0 +1,61 @@
...
+            json.getNames().asScala.map(name => UserGreeting(name, json.getString(name))).toList
+          case None => List.empty[UserGreeting]
+        }
+      }
+  }
+  // #service-impl
+
+  //#register-event-processor
+  ???
+  //#register-event-processor
```
johanandren: @ygree if you know what should be here and can point me in the right direction that would be great

docs/src/test/scala/docs/home/persistence/CouchbaseReadSideProcessor.scala
ygree: No, CouchbaseReadSide shouldn't extend ReadSide despite the naming. ReadSide is the register that usually used in the application loader to register read-side processors, e.g. `readSide.register(wire[HelloEventProcessor])`

docs/src/test/scala/docs/home/persistence/CouchbaseReadSideQuery.scala
ygree:
```
  class HelloServiceImpl(
    persistentEntityRegistry: PersistentEntityRegistry,
    readSide: ReadSide,
    couchbase: CouchbaseSession) extends HelloService {

    readSide.register[HelloEvent](new HelloEventProcessor(couchbase))
```

docs/src/test/scala/docs/home/persistence/CouchbaseReadSideProcessor.scala
ygree: Perhaps, `ReadSide` should have been called `ReadSideRegistry`

PR #153, opened 2018-12-13
Latest Akka and Lagom RC

johanandren:


chbatey:
The latest Lagom RC has broken Whitresource for bringing in:

```
[error] java.lang.RuntimeException: Some dependencies were rejected by the organization's policies:
[error] akka-persistence-couchbase-master:
[error] * javax.activation-api-1.2.0.jar (rejected by policy '*GPL* - Reject')
```

Guess that should be provided

PR #159, opened 2018-12-17
Alpakkify a bit more

ennru:
* Use java.util.Objects for hashCode and equals
* talkative toString 
* use sbt's `Test` for scoping


couchbase-client/src/main/scala/akka/stream/alpakka/couchbase/settings.scala
```diff
@@ -128,17 +135,21 @@ final class CouchbaseSessionSettings private (val username: String,
...
   override def equals(other: Any): Boolean = other match {
     case that: CouchbaseSessionSettings =>
-      username == that.username &&
-      password == that.password &&
-      nodes == that.nodes &&
-      environment == that.environment
+      java.util.Objects.equals(this.username, that.username) &&
+      java.util.Objects.equals(this.password, that.password) &&
+      java.util.Objects.equals(this.nodes, that.nodes) &&
+      java.util.Objects.equals(this.environment, that.environment)
```
johanandren: Not that I'm against it, but, why do you guys do this when Scala `==` is null safe?

project/Dependencies.scala
```diff
@@ -71,6 +71,8 @@ object Dependencies {
     couchbaseDcpClient,
     rxJavaReactiveStreams,
     scalaTest,
+    "com.typesafe.play" %% "play-json" % "2.6.9" % Test, // MIT like: http://www.slf4j.org/license.html
+    "com.typesafe.akka" %% "akka-http" % "10.1.5" % Test, // MIT like: http://www.slf4j.org/license.html
```
johanandren: Why do we need to depend on these?

Issue #113, opened 2018-12-04
Semantics around events written after currentEventsByTag


johanandren:
It's a bit unclear if `currentEventsByTag` includes new events with a time based uuid further back in time than the consistency-delay when they are written after the query was executed/started. We should either have a test or note in the docs it is not quite defined.

johanandren:
Strange thing is: with the current query and paging I don't see how new events can be filtered out, however I tried to reproduce this in a test in #118 but I couldn't get it to fail.

chbatey:
I had a look at how the C* plugin does this.

* It does use eventual consistency delay for current queries
* It refreshes the query until it is at current time (set at toOffset) + eventual consistency delay by always staying behind by the eventual consistency delay and stopping when it hits the toOffset

This means that new writes that are written after the the start of the current query but have an offset than less than when the currentQuery was started will be delivered.

chbatey:
Couchbase doesn't limit the query by the offset so this isn't going to work the same here.

The reason the test fails in #118 is that the stage gets demand of more than the probe requested, I assume due to the mapAsync we do for its buffer

johanandren:
A bit of confusion here from me, sorry, the problem was about events by persistence id.

I had a call with Chris and we decided that we should solve this by starting out with a query to find out the highest sequence number at the time the query is triggered and then limit the actual events query using that.

PR #162, opened 2018-12-17
CurrentEventsByPersistenceId: Stop at the current highest seq nr

chbatey:
Refs #113


core/src/main/scala/akka/persistence/couchbase/scaladsl/CouchbaseReadJournal.scala
```diff
@@ -286,7 +299,7 @@ final class CouchbaseReadJournal(eas: ExtendedActorSystem, config: Config, confi
             // offset delay ms back in time from now
             TimeBasedUUIDs.create(
               UUIDTimestamp.fromUnixTimestamp(
-                System.currentTimeMillis() - (settings.eventByTagSettings.eventualConsistencyDelay.toMillis)
+                System.currentTimeMillis() - settings.eventByTagSettings.eventualConsistencyDelay.toMillis
```
johanandren: heh, that was weird. Did I put those there?

core/src/test/scala/akka/persistence/couchbase/scaladsl/EventsByPersistenceIdSpec.scala
```diff
@@ -123,47 +123,52 @@ class EventsByPersistenceIdSpec extends AbstractQuerySpec("EventsByPersistenceId
...
           .runWith(Sink.seq)
           .futureValue
-        evts.map(_.event) should ===(List(s"$pid-$seqNrAfterInitial"))
+        events.map(_.event) should ===(List(s"$pid-$seqNrAfterInitial"))
       }
 
       // here comes the weird part, the query stage should fill up one page and not do another query until
       // we reach the end of it, but then that second query should see the new event and fail this test
       // but it doesn't
+      val expectedEvents = ((notTheEntireFirstPage + 1L) to initialPersistedEvents).map(n => s"$pid-$n")
```
johanandren: the comment is not valid any more

core/src/main/scala/akka/persistence/couchbase/scaladsl/CouchbaseReadJournal.scala
```diff
@@ -155,6 +151,9 @@ final class CouchbaseReadJournal(eas: ExtendedActorSystem, config: Config, confi
                                      toSequenceNr: Long): Source[EventEnvelope, NotUsed] =
     internalEventsByPersistenceId(live = true, persistenceId, fromSequenceNr, toSequenceNr)
 
+  // Should this be configurable?
+  private val queryConsistency = N1qlParams.build().consistency(ScanConsistency.REQUEST_PLUS)
```
johanandren: I don't think this shouldn't be the default here, it's expensive, and only really needed on the journal side for replay (see all tests counting on eventual consistency for the index).

core/src/main/scala/akka/persistence/couchbase/scaladsl/CouchbaseReadJournal.scala
chbatey: Think an elf did

core/src/test/scala/akka/persistence/couchbase/scaladsl/EventsByPersistenceIdSpec.scala
chbatey: :+1:

core/src/main/scala/akka/persistence/couchbase/scaladsl/CouchbaseReadJournal.scala
chbatey: :+1: current has no real guarantees

johanandren:
Two failures, one is unrelated, but the other one seems relevant: 
`EventsByPersistenceId must find existing events`

chbatey:
Tests are failing because the query to get the sequenceNr is only executed once

johanandren:
I can take this the last few steps to get it in another RC today.

PR #163, opened 2018-12-17
[WIP] Transposing JSON AST:s between Jackson and Couchbase

johanandren:
Just a spike on how we could potentially do #76 

Not sure it is worth duplicating that (one more time for play-json) and then also the migration logic from the Lagom Jackson module would need to be duplicated.

### DIFF
--- a/lagom-persistence-couchbase/javadsl/src/main/scala/com/lightbend/lagom/javadsl/persistence/couchbase/JacksonCouchbaseSerializer.scala
+++ b/lagom-persistence-couchbase/javadsl/src/main/scala/com/lightbend/lagom/javadsl/persistence/couchbase/JacksonCouchbaseSerializer.scala
@@ -1,0 +1,163 @@
+/*
+ * Copyright (C) 2018 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package com.lightbend.lagom.javadsl.persistence.couchbase
+import java.io.{InputStream, NotSerializableException}
+import java.math.BigInteger
+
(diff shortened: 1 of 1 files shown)


lagom-persistence-couchbase/javadsl/src/main/scala/com/lightbend/lagom/javadsl/persistence/couchbase/JacksonCouchbaseSerializer.scala
```diff
@@ -0,0 +1,163 @@
...
+    val objectNode = nodeFactory.objectNode()
+    objectNode.setAll(fields)
+  }
+}
+
+class JacksonCouchbaseSerializer(as: ExtendedActorSystem) extends JsonSerializer {
+
+  import JacksonCouchbaseSerializer._
+
+  // FIXME if this was to be actually used we need the same kind of migration infra as the JacksonJsonSerializer from lagom
```
patriknw: I'm somewhat worried that we introduce a maintenance burden by providing this, given that we want to have Jackson serialization in Akka and eventually replace corresponding in Lagom with that one. Might be better to wait until it has been asked for.

johanandren: I agree, was mostly curious about what APIs for dealing with a JSON AST there is on Jackson as opposed to going via json rendered to a string.

johanandren:
Closing this for now as it was just "for show"

PR #168, opened 2018-12-18
Dont call watch from future callback

chbatey:


chbatey:
Actually i think the watch is more broken, looking ingo it

PR #174, opened 2018-12-19
Keeping track of the persistent actors fixed

### LINKED ISSUE #169: Watching of persistent actors doesn't work
`asyncReplayMessages` is not called on the actor thread so we can't get the persistent actor with `sender()`. Also replay is not always called. E.g. on the first time the persistenceId exists and if there is a current snapshot.

We need to do it in `asyncReadHighestSequenceNr` which is called on the journal thread when receiving a message from `Eventsourced`

johanandren:
Fixes #169


johanandren:
Merging this without review because of releasing soon and having to wait for an additional slow travis validation that needs to happen before we can release

PR #178, opened 2019-01-03
Events by tag covering index for id and ordering

chbatey:
Then use the KV API to get the payloads


core/src/main/scala/akka/persistence/couchbase/internal/CouchbaseSchema.scala
```diff
@@ -108,12 +108,12 @@ private[akka] final object CouchbaseSchema {
...
-         |AND m.sequence_nr >= $$from
-         |AND m.sequence_nr <= $$to
-         |ORDER BY m.sequence_nr
+         SELECT a.persistence_id, a.writer_uuid, m.* FROM ${bucketName} a UNNEST messages AS m
+         WHERE a.type = "${CouchbaseSchema.JournalEntryType}"
+         AND a.persistence_id = $$pid
+         AND m.sequence_nr >= $$from
+         AND m.sequence_nr <= $$to
+         ORDER BY m.sequence_nr
     """.stripMargin
```
patriknw: .stripMargin not needed

core/src/main/scala/akka/persistence/couchbase/scaladsl/CouchbaseReadJournal.scala
```diff
@@ -295,29 +292,50 @@ final class CouchbaseReadJournal(eas: ExtendedActorSystem, config: Config, confi
...
+          val ordering = value.getString("ordering")
+          session.get(id).map {
+            case Some(doc) => (ordering, doc)
+            case None =>
+              throw new IllegalStateException(
+                s"Document from events by tag index not found. (concurrent deletes?). Id: $id"
+              )
+          }
+        }
+        .map { orderingAndDoc =>
```
patriknw:
```
case (ordering, doc) =>
```

core/src/main/scala/akka/persistence/couchbase/scaladsl/CouchbaseReadJournal.scala
```diff
@@ -295,29 +292,50 @@ final class CouchbaseReadJournal(eas: ExtendedActorSystem, config: Config, confi
...
+          val specificDoc = ((0 until messages.size) map { i =>
+            messages.getObject(i)
+          }).find(_.getString(CouchbaseSchema.Fields.Ordering) == ordering) match {
+            case Some(d) => d
+            case None =>
+              throw new IllegalStateException(s"Expected to find doc ${doc} to have message with ordering ${ordering}")
+          }
+          (persistenceId, specificDoc)
+        }
+        .mapAsync(1) { row =>
```
patriknw: `case` for the row parts

docker/couchbase-1-node/configure.sh
```diff
@@ -53,8 +53,7 @@ sleep 20 #TODO: how to check if it's ready to process N1QL queries
 log "Create bucket indices ........."
 cbq -u $USERNAME -p $PASSWORD -s "CREATE INDEX \`persistence-ids\` on \`akka\` (\`persistence_id\`) WHERE \`type\` = \"journal_message\""
 cbq -u $USERNAME -p $PASSWORD -s "CREATE INDEX \`sequence-nrs\` on \`akka\` (DISTINCT ARRAY m.sequence_nr FOR m in messages END) WHERE \`type\` = \"journal_message\""
-cbq -u $USERNAME -p $PASSWORD -s "CREATE INDEX \`tags\` ON \`akka\` (ALL ARRAY m.tags FOR m IN messages END) WHERE \`type\` = \"journal_message\""
-cbq -u $USERNAME -p $PASSWORD -s "CREATE INDEX \`tags-ordering\` ON \`akka\` (DISTINCT ARRAY m.ordering FOR m IN messages END) WHERE \`type\` = \"journal_message\""
+cbq -u $USERNAME -p $PASSWORD -s "CREATE INDEX \`tags-ordering-cover-meta\` ON \`akka\`((ALL (ARRAY (ALL (ARRAY [\`t\`, (\`m\`.\`ordering\`)] FOR \`t\` IN (\`m\`.\`tags\`) END)) FOR \`m\` IN \`messages\` END))) WHERE (\`type\` = \"journal_message\")"
```
johanandren: should we maybe just call it "tags" instead?

core/src/main/scala/akka/persistence/couchbase/scaladsl/CouchbaseReadJournal.scala
```diff
@@ -295,29 +292,49 @@ final class CouchbaseReadJournal(eas: ExtendedActorSystem, config: Config, confi
...
-        .mapAsync(1) { row: AsyncN1qlQueryRow =>
-          val json = row.value()
-          CouchbaseSchema
-            .deserializeTaggedEvent(json, Long.MaxValue, serialization)
+      taggedRowIds
+        .mapAsync(settings.eventByTagSettings.getParallelism) { row =>
+          val value = row.value()
+          val id = value.getString("id")
+          val ordering = value.getString("ordering")
+          session.get(id).map {
```
johanandren: I guess this means that we will fetch the same document several time for a batched write? Can't think of a way to avoid doing that which isn't overly complicated though (db cleverness - ordering becomes a problem, cache - when do we invalidate, needs to be threadsafe etc).

core/src/main/scala/akka/persistence/couchbase/scaladsl/CouchbaseReadJournal.scala
```diff
@@ -295,29 +292,49 @@ final class CouchbaseReadJournal(eas: ExtendedActorSystem, config: Config, confi
...
+        .map { case (ordering, doc) =>
+          val persistenceId = doc.content().getString(CouchbaseSchema.Fields.PersistenceId)
+          val messages = doc.content().getArray("messages")
+          val specificDoc = ((0 until messages.size) map { i =>
+            messages.getObject(i)
+          }).find(_.getString(CouchbaseSchema.Fields.Ordering) == ordering) match {
+            case Some(d) => d
+            case None =>
+              throw new IllegalStateException(s"Expected to find doc ${doc} to have message with ordering ${ordering}")
+          }
```
johanandren: This looks weird, isn't there a Java collection so we can do a `getArray().asScala.find(...)` instead of the range map?

core/src/main/scala/akka/persistence/couchbase/scaladsl/CouchbaseReadJournal.scala
```diff
@@ -295,29 +292,49 @@ final class CouchbaseReadJournal(eas: ExtendedActorSystem, config: Config, confi
...
+            case None =>
+              throw new IllegalStateException(
+                s"Document from events by tag index not found. (concurrent deletes?). Id: $id"
+              )
+          }
+        }
+        .map { case (ordering, doc) =>
+          val persistenceId = doc.content().getString(CouchbaseSchema.Fields.PersistenceId)
+          val messages = doc.content().getArray("messages")
+          val specificDoc = ((0 until messages.size) map { i =>
```
johanandren: nitpick: not a `doc` here, just a JsObject

docker/couchbase-1-node/configure.sh
chbatey: Yep, will rename

core/src/main/scala/akka/persistence/couchbase/scaladsl/CouchbaseReadJournal.scala
chbatey: Yes that's right and agreed that it would add too much complexity to avoid it

PR #180, opened 2019-01-04
Remove warnings + make them fatal

chbatey:
We had so few I thought we could make them fatal until we have a reason to remove it


chbatey:
ah root doesn't aggregate them all, guess i have more warnings to fix!

chbatey:
There were a few more than originally thought but still worth it

PR #184, opened 2019-01-07
Max tag sequence number query/index that scales

johanandren:


core/src/main/scala/akka/persistence/couchbase/internal/CouchbaseSchema.scala
```diff
@@ -228,15 +228,21 @@ private[akka] final object CouchbaseSchema {
...
-        AND a.persistence_id = $$pid
-        AND t.tag = $$tag
-        ORDER BY m.sequence_nr DESC
-        LIMIT 1
+         SELECT [a.persistence_id, t.tag, t.seq_nr][2] as seq_nr
+         FROM ${bucketName} AS a USE INDEX (`tag-seq-nrs`)
+         UNNEST a.messages AS m
+         UNNEST m.tag_seq_nrs AS t
+         WHERE a.type = "${CouchbaseSchema.JournalEntryType}"
+         AND [a.persistence_id, t.tag, t.seq_nr] >= [$$pid, $$tag, SUCCESSOR(0)]
```
chbatey: 1 instead of successor(0)?

PR #189, opened 2019-01-09
Avoid repeating the tag info twice in the stored json

### LINKED ISSUE #176: Eliminate duplication of tag
In tagged events we put both
```
tags: ["tag-1", ...]
```
and 
```
tag_seq_nrs: [{ tag: "tag-1", seqnr: 1 }, ...]
```

This was originally to be able to index the first one, but I think we could pull it off with just the second representation instead.

johanandren:
Fixes #176


johanandren:
Renaming tag_seq_nrs to tags and using that for both the events by tag and the out-of-order detection logic

core/src/main/scala/akka/persistence/couchbase/internal/CouchbaseSchema.scala
```diff
@@ -147,14 +150,14 @@ private[akka] final object CouchbaseSchema {
...
-         SELECT meta(a).id, [t, m.ordering][1] as ordering
+         SELECT meta(a).id, [t.tag, m.ordering][1] as ordering
          FROM ${bucketName} a USE INDEX (tags)
          UNNEST messages m
          UNNEST m.tags t
          WHERE a.type = "${CouchbaseSchema.JournalEntryType}"
-         AND [t, m.ordering] >= [$$tag, SUCCESSOR($$fromOffset)]
-         AND [t, m.ordering] <= [$$tag, $$toOffset]
-         ORDER BY [t, m.ordering]
+         AND [t.tag, m.ordering] >= [$$tag, SUCCESSOR($$fromOffset)]
```
chbatey: did you validate that this still pushes down the limit?

johanandren: No I think I committed this by accident to be honest, damn

johanandren: Ach, misread the change, it should be ok, I'll double check anyways.

johanandren: Double checked that the explain output looks the same! 👍

PR #190, opened 2019-01-11
Use j.u.c.Executor instead of ExecutionContext in Java API

ennru:
The Java API should not expose Scala classes.


couchbase-client/src/main/scala/akka/stream/alpakka/couchbase/javadsl/CouchbaseSession.scala
```diff
@@ -32,9 +32,9 @@ object CouchbaseSession {
    */
   def create(settings: CouchbaseSessionSettings,
              bucketName: String,
-             executionContext: ExecutionContext): CompletionStage[CouchbaseSession] =
+             executor: Executor): CompletionStage[CouchbaseSession] =
     ScalaDslCouchbaseSession
-      .apply(settings, bucketName)(executionContext)
+      .apply(settings, bucketName)(ExecutionContext.fromExecutor(executor))
```
patriknw: Akka dispatchers are already `ExecutionContext` so it might be more optimal to do a match on that first:
```
executor match {
  case ec: ExecutionContext => ec
  case _ => ExecutionContext.fromExecutor(executor)
}
```

I don't think that is done in `ExecutionContext.fromExecutor`

PR #194, opened 2019-02-25
Update Lagom to version 1.5.0-RC1

marcospereira:
The primary purpose here is to update Lagom to version 1.5.0-RC1. But while I'm here, there are some more updates too.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -15,7 +15,7 @@ def common: Seq[Setting[_]] = Seq(
                           "https://gitter.im/akka/dev",
                           url("https://github.com/akka/akka-persistence-couchbase/graphs/contributors")),
   licenses := Seq(("Apache-2.0", url("http://www.apache.org/licenses/LICENSE-2.0"))),
-  crossScalaVersions := Seq("2.12.7", "2.11.12"),
+  crossScalaVersions := Seq("2.12.8", "2.11.12"),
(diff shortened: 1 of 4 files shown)


johanandren:
I was discussing per mail with Tim if we shouldn't rather change the Lagom version to 1.4.12 for the 1.0 release to not be blocked on the Lagom release. Haven't looked into that yet though.

TimMoore:
I don't have a strong opinion on whether it is compatible with 1.4.x or not, I think we should just do whatever is easier and presents the fewest obstacles.

TimMoore:
@marcospereira could you update to RC2? I don't seem to have permission to update your branch.

johanandren:
Coordinated shutdown APIs changed?

```
[error] /home/travis/build/akka/akka-persistence-couchbase/lagom-persistence-couchbase/scaladsl/src/test/scala/com/lightbend/lagom/scaladsl/persistence/couchbase/PersistentEntityRefSpec.scala:81:56: object creation impossible, since method coordinatedShutdown in trait AkkaManagementComponents of type => akka.actor.CoordinatedShutdown is not defined
[error]   val components: CouchbasePersistenceComponents = new CouchbasePersistenceComponents {
[error]                                                        ^
[error] one error found
```

PR #196, opened 2019-03-05
Upgrade to sbt 1.2.8, and latest plugin versions

ennru:


project/build.properties
```diff
@@ -1 +1 @@
-sbt.version=1.2.6
+sbt.version=1.2.7
```
2m: `1.2.8` is the latest

PR #201, opened 2019-03-13
JDK 11 / unbreak travis in general

johanandren:


project/Dependencies.scala
```diff
@@ -6,14 +6,15 @@ import sbt._
 
 object Dependencies {
 
-  val AkkaVersion = "2.5.21"
+  val AkkaVersion = "2.5.20"
```
raboof:
```suggestion
  val AkkaVersion = "2.5.21"
```

lagom-persistence-couchbase/javadsl/src/main/scala/com/lightbend/lagom/javadsl/persistence/couchbase/CouchbasePersistenceModule.scala
```diff
@@ -84,7 +83,6 @@ private[lagom] object CouchbasePersistenceModule {
   class InitServiceLocatorHolder @Inject()(system: ActorSystem, injector: Injector) {
 
     // Guice doesn't support this, but other DI frameworks do.
```
raboof:
```suggestion
```

raboof: Lagom fixes this by adding the `"javax.annotation" % "javax.annotation-api" % "1.3.2"` dependency - might be worth considering here as well, instead of removing the annotations?

johanandren: Strange that it is not brought in transitively then. But if it is anyways not supported by Guice I don't get why we'd bother.

raboof: It's not merged there yet (https://github.com/lagom/lagom/pull/1780/files#diff-0ecdbc5a001d52fb34f5eafb7cd1aaa6R356). As far as I understand you can use Play/Lagom with other DI frameworks then Guice, even though that's not used a lot.

PR #206, opened 2019-04-16
Lagom 1.5.0

renatocaval:


johanandren:
I think it was temporary failures, retriggered those two build steps that failed now.

TimMoore:
Looks like it could be a legitimate compatibility problem. Both jobs failed again with:

```
Running java with options -classpath [...lots of jars...] sbt.ForkMain 37540 failed with exit code 255
```

I get the same when I checkout the PR branch and run `sbt lagom-persistence-couchbase-scaladsl/test` locally.

TimMoore:
Oh, I see, that part is just a generic "some tests failed" error (funny I have never seen it in that form before).

TimMoore:
I don't see any other errors on Travis, but the ones I get locally look like this:

```
[info]   java.lang.AssertionError: assertion failed: expected Appended(A), found com.lightbend.lagom.scaladsl.persistence.PersistentEntity$PersistException: Persist of [akka.persistence.journal.Tagged] failed in [com.lightbend.lagom.scaladsl.persistence.TestEntity] with id [1], caused by: {No index available on keyspace akka that matches your query. Use CREATE INDEX or CREATE PRIMARY INDEX to create an index, or check that your expected index is online.
```

Docker Compose ran successfully and it looks like all of the expected indexes were created successfully in the container log. I'll have to leave it to someone who knows what they're doing to diagnose further.

johanandren:
I get the same locally, but no the actual assertion error Tim mentions.

johanandren:
I think the problem is something related to shutdown order in `PersistentEntityRefSpec`, if I switch order of the call to `super.afterAlll` and stopping the actor system in `afterAll` I get a different error.

```
[error] Uncaught exception when running tests: java.lang.IllegalStateException: done method is called twice
[error] sbt.ForkMain$ForkError: java.lang.IllegalStateException: done method is called twice
[error] 	at org.scalatest.tools.Framework$ScalaTestRunner.done(Framework.scala:778)
[error] 	at sbt.ForkMain$Run.runTests(ForkMain.java:266)
[error] 	at sbt.ForkMain$Run.run(ForkMain.java:144)
[error] 	at sbt.ForkMain$Run.access$100(ForkMain.java:140)
[error] 	at sbt.ForkMain.main(ForkMain.java:126)
```

Could it be related to graceful shutdown, has something around shutting down changed since RC1?

TimMoore:
@ignasi35 do you have any thoughts about this?

johanandren:
Could it be something in `copy-of-lagom-persistence-test` that was copied from M1, that worked in RC1 but not anymore in GA?

TimMoore:
It was passing with RC2, right? Here are the diffs: https://github.com/lagom/lagom/compare/1.5.0-RC2...1.5.0

My suspicion: https://github.com/lagom/lagom/compare/1.5.0-RC2...1.5.0#diff-7872453b725d07dfe4dedc2e195f9ffcR18

This probably needs to be overridden back to `off` for tests.

renatocaval:
I disabled the exit-jvm on tests now. 

Let's see if it passes.

ignasi35:
I finally got the TODO empty. Looking...

johanandren:
Yeah, tests are a bit flaky on travis I'm afraid. Restarted.

ignasi35:
Flaky tests aside, there's a real issue on the `scaladsl` tests.

I've run the tests locally disabling `fork`ing and, while `lagom-persistence-couchbase-javadsl/test` complete normally, `lagom-persistence-couchbase-scaladsl/test` causes the `sbt` process to exit. 

I'm enabling Akka and Lagom DEBUG logs to get a better insight at what's the actual cause.

renatocaval:
Possible cause may be another missing property, 

```
lagom.cluster.bootstrap.enabled = off
```

This is needed in tests that start the cluster programmatically. That doesn't explain why the javadsl is passing though.

I'm testing it locally to see if I can trace it.

renatocaval:
Couldn't solve it yet. To be continued tomorrow...

ignasi35:
The problem was not obvious because the setting `lagom.cluster.exit-jvm-when-system-terminated = off` is [only used on a Coordinated Shutdown task](https://github.com/lagom/lagom/blob/9b050f32b38ece1d00d22f0ded442e7c6032a4de/cluster/core/src/main/scala/com/lightbend/lagom/internal/cluster/JoinClusterImpl.scala#L78-L122). But the test causing the JVM to exit is [not using Coordinated Shutdown](https://github.com/akka/akka-persistence-couchbase/blob/ff334af2e7f3026f7654e1a10e8e2aa80b57862f/lagom-persistence-couchbase/scaladsl/src/test/scala/com/lightbend/lagom/scaladsl/persistence/couchbase/PersistentEntityRefSpec.scala#L67).

Well, it turns out, that using `actorSystem.terminate` on an actor system that is clustered [will invoke Coordinated Shutdown](https://github.com/akka/akka/blob/8dd6f790bd5b939e9419dc5d75fd70753b7d85b1/akka-cluster/src/main/scala/akka/cluster/ClusterDaemon.scala#L212-L217). And not only that, it will be invoked with a `ClusterDowningReason`.

I think the test code (both here and in Lagom itself, since this particular test is almost a verbatim copy of Lagom's [`PersistenceEntityRefSpec`](https://github.com/lagom/lagom/blob/cc7dbf595b6af72a20c685d153fe3cb7f8bcae53/persistence-cassandra/scaladsl/src/test/scala/com/lightbend/lagom/scaladsl/persistence/cassandra/PersistentEntityRefSpec.scala#L35)) should not use `Testkit.shutdownActorSystem(system)` and opt into using `CoordinatedShutdown`. I suspect that, in these particular tests, `Testkit.shutdownActorSystem(system)` is hiding the use of Coordinated Shutdown and implementing the termination in terms of `CoordinatedShutdown` would be clearer.

renatocaval:
@ignasi35, I agree. But even when we move to explicit finalizing the set with a call to CoordinatedShutdown, we still need to set the property `lagom.cluster.exit-jvm-when-system-terminated` to `off`. 

And the `PersistenceEntityRefSpec` is using its own set of cluster config. If it was using the Lagom config like in the other tests, we would not have hit that issue. 

Let's change that on another PR and also be explicit about finalizing the test with coordinated shutdown (both here as in Lagom).

lagom-persistence-couchbase/core/src/test/scala/com/lightbend/lagom/internal/persistence/couchbase/TestConfig.scala
```diff
@@ -17,7 +17,8 @@ object TestConfig {
     akka.cluster.sharding.distributed-data.durable.keys: []
     lagom.cluster.join-self: on
     lagom.cluster.bootstrap.enabled: off
-    lagom.akka.management.enabled: off
+    lagom.akka.management.enabled: off,
+    lagom.cluster.exit-jvm-when-system-terminated: off
```
renatocaval: I didn't get the reason why we have an apart cluster config. Most tests are not using it, only `PersistentEntityRefSpec`. That's probably a left-over.

We can change it on another PR. I would like it to pass the build and get merged for now.

ignasi35:
> @ignasi35, I agree. But even when we move to explicit finalizing the set with a call to CoordinatedShutdown, we still need to set the property `lagom.cluster.exit-jvm-when-system-terminated` to `off`.

We don't. :-)

When replacing `Testkit.shutdownActorSystem(system)` with `coordinatedShutdown.run(...)` the developer must provide a `Reason` and only if the `Reason` is on the [whitelist](https://github.com/lagom/lagom/blob/9b050f32b38ece1d00d22f0ded442e7c6032a4de/cluster/core/src/main/scala/com/lightbend/lagom/internal/cluster/JoinClusterImpl.scala#L83-L85) would Lagom try to exit the JVM.

Hmm, the more I think about this approach the more I like it. Using `coordinatedShutdown.run(...)` makes the use of the setting override unnecessary. We could eventually remove the override from all Lagom tests.

.travis.yml
```diff
@@ -98,4 +98,4 @@ env:
...
+    - secure: "MKFv2AAOhY4l7NCZY7Wa2OCiM8eL7LKpCRbF+K96YV3F+lHNCQIwTrYY8B9eprI/WbS+s6KszU8hVIdaxVxATK+k7DRYAG/54NkBVMx2wPEBupMovdvtEiUmFXIacAH6qY3lv0JeTvCu1tq1G/EEdNEgl51T0wjstcJWQEzqnwUOc/GytPsN0vM79lntNAR5PZ67HTDwW30i/xgsUHU1jYqTVFFZLXoAjkxVJ+Yub+lN2+lFHb2VsG971SQ8GO2bUANYlSelr6cF7k7B3J0au5lueHDDdOGUhsDSrGVmjZq5Ph0KU97AhvuyMHNdfjvwBdVR1i/lOl4f3rSjdxpgnUFCLbaQIj3kZ5vSOHY4dDTbYzKvj1fgf9GRB5g1t/TnQZEkziTyvs/BQklmAV0oM6KzUXHVUtvp6IEiqcko5aeye94naX/YCmZuMpiWCiAgAfqP+DTpufA3yMSWIZOsUF8pi+7tnlVcAiDvh6UOygRPx5jONtz5KBp9FD9ISU4yKyPdILHTtZd1BnuD+/r7AoJTux3qu5ukShF+8PFXncd4HooTTnw3euH2LFt8RYmqQp01DOBj2O0Fl+OjXJWvFRSKjAdLRkXroOEHR8HPbBq9ow0TVOtE7nqvajoymEAx77N3omcA27kOXHQL1uxnLZxIckpHRK4pp74fSiyPins="
```
johanandren: Why did this need a change?

ignasi35: I think it's just formatting (remove an extra blankspace)

PR #209, opened 2019-04-24
-m

renatocaval:


lagom-persistence-couchbase/core/src/main/scala/com/lightbend/lagom/internal/persistence/couchbase/CouchbaseReadSideSessionProvider.scala
```diff
@@ -1,19 +0,0 @@
...
-
-import scala.concurrent.ExecutionContext
-
-/**
- * Internal API
- */
-private[lagom] object CouchbaseReadSideSessionProvider {
-
-  def apply(system: ActorSystem, executionContext: ExecutionContext): CouchbaseSession =
-    ???
```
renatocaval: It's clear that this one is not being used. Probably a leftover from some previous refactoring.

renatocaval:
Sorry for this stupid commit message and PR title. I didn't see it. I have a script that commits a change and sends a PR in one shot and apparently, I did some bad manipulation and it used `-m` as the message.

Issue #211, opened 2019-04-25
Prefer x.y.z version names


ignasi35:
Prior to 1.0.0, it is common in the Akka ecosystem to use `0.x` versions.

We should converge to `x.y.z` after `1.0.0

patriknw:
I agree, it should be 1.0.x. Let’s fix for 1.0.1

johanandren:
Ah, sorry, I didn't think about this sine we started out with a two digit version.

ennru:
This uses the major.minor.patch scheme now.

PR #213, opened 2019-04-29
Update project support info for docs

johanandren:


johanandren:
Snuck in a pre-fix for #211 here as well

johanandren:
Enno pointed out that we use sbt-dynver here, so no way to pre-fix #211

PR #221, opened 2019-07-11
Lagom 1.6.0-M4

### LINKED ISSUE #204: Publish Lagom modules for Scala 2.13
Depends on lagom/lagom#1240

marcospereira:
## Purpose

Fixes #204.

Update to use Lagom 1.6.0-M4. This version is cross built to Scala 2.13, so this PR also updates from 2.13.0-M5 to 2.13.0 GA, which then requires some other deps updates.

Since `scalacOptions` is configured to fail on warnings, some code changes were made.

## Changes

- Lagom 1.6.0-M4
- Scala 2.13
- Akka 2.6.0-M4 (meaning Scala 2.11 is gone).

## Missing tasks

- [ ] Travis job to build using Scala 2.13.0

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -34,7 +34,6 @@ def common: Seq[Setting[_]] = Seq(
     "-Xlint",
     "-Ywarn-dead-code",
     "-Ywarn-numeric-widen",
-    "-Xfuture",
     "-Xfatal-warnings"
   ),
   bintrayOrganization := Some("akka"),
@@ -128,7 +127,6 @@ lazy val `copy-of-lagom-persistence-test` =
       // This modules copy-pasted preserve it as is
       scalafmtOnCompile := false,
       libraryDependencies := Dependencies.`copy-of-lagom-persistence-test`,
-      crossScalaVersions -= Dependencies.Scala213,
     )
 
 lazy val `lagom-persistence-couchbase-core` = (project in file("lagom-persistence-couchbase/core"))
@@ -139,7 +137,6 @@ lazy val `lagom-persistence-couchbase-core` = (project in file("lagom-persistenc
   .settings(
     name := "lagom-persistence-couchbase-core",
     libraryDependencies := Dependencies.`lagom-persistence-couchbase-core`,
-    crossScalaVersions -= Dependencies.Scala213,
   )
 
 lazy val `lagom-persistence-couchbase-javadsl` = (project in file("lagom-persistence-couchbase/javadsl"))
@@ -154,7 +151,6 @@ lazy val `lagom-persistence-couchbase-javadsl` = (project in file("lagom-persist
   .settings(
     name := "lagom-javadsl-persistence-couchbase",
     libraryDependencies := Dependencies.`lagom-persistence-couchbase-javadsl`,
-    crossScalaVersions -= Dependencies.Scala213,
   )
   .enablePlugins(MultiJvmPlugin)
   .configs(MultiJvm)
@@ -172,7 +168,6 @@ lazy val `lagom-persistence-couchbase-scaladsl` = (project in file("lagom-persis
   .settings(
     name := "lagom-scaladsl-persistence-couchbase",
     libraryDependencies := Dependencies.`lagom-persistence-couchbase-scaladsl`,
-    crossScalaVersions -= Dependencies.Scala213,
   )
   .enablePlugins(MultiJvmPlugin)
   .configs(MultiJvm)
--- a/build.sbt
+++ b/build.sbt
@@ -21,7 +21,7 @@ inThisBuild(
 )
 
 def common: Seq[Setting[_]] = Seq(
-  crossScalaVersions := Seq(Dependencies.Scala213, Dependencies.Scala212, Dependencies.Scala211),
+  crossScalaVersions := Seq(Dependencies.Scala213, Dependencies.Scala212),
   scalaVersion := Dependencies.Scala212,
   crossVersion := CrossVersion.binary,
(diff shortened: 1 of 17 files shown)


build.sbt
```diff
@@ -34,7 +34,6 @@ def common: Seq[Setting[_]] = Seq(
     "-Xlint",
     "-Ywarn-dead-code",
     "-Ywarn-numeric-widen",
-    "-Xfuture",
```
marcospereira: Deprecated in Scala 2.13, so I removed it. But we can add it conditionally if you consider it is a better option.

core/src/test/scala/akka/testkit/WithLogCapturing.scala
```diff
@@ -16,7 +16,7 @@ import org.scalatest.{Outcome, SuiteMixin, TestSuite}
 /**
  * Capture all log events and print them only for failing tests
  */
-trait WithLogCapturing extends SuiteMixin { this: TestSuite ⇒
```
marcospereira: `⇒` is deprecated in Scala 2.13.

core/src/main/scala/akka/persistence/couchbase/internal/N1qlQueryStage.scala
```diff
@@ -109,7 +109,7 @@ private[akka] final class N1qlQueryStage[S](live: Boolean,
 
       override def preStart(): Unit = {
         if (live)
-          schedulePeriodicallyWithInitialDelay(Poll, settings.liveQueryInterval, settings.liveQueryInterval)
+          scheduleAtFixedRate(Poll, settings.liveQueryInterval, settings.liveQueryInterval)
```
marcospereira: `schedulePeriodicallyWithInitialDelay` is deprecated.

lagom-persistence-couchbase/scaladsl/src/test/scala/com/lightbend/lagom/scaladsl/persistence/couchbase/CouchbaseReadSideSpec.scala
```diff
@@ -40,9 +39,7 @@ class CouchbaseReadSideSpec
 
   override def getAppendCount(id: String): Future[Long] = readSide.getAppendCount(id)
 
-  override def afterAll(): Unit = {
-    persistentEntityRegistry.gracefulShutdown(5.seconds)
```
marcospereira: `gracefulShutdown` was a no-op in Lagom 1.5.x and it was removed in 1.6.x.

marcospereira:
@johanandren maybe this has too many changes, but I kept them isolated in smaller commits. I'm also unsure about the policy around updating Akka and dropping Scala 2.11 for these satellite projects.

core/src/main/scala/akka/persistence/couchbase/internal/N1qlQueryStage.scala
johanandren: I wonder if we backported that to 2.5 though?

johanandren:
Tricky, having to drop 2.11 is only a consequence of having the Lagom support in the same release-build as the core Couchbase journal stuff, right? Could be the time we need to split the two so that the journal can stay compatible with Akka 2.5 + Scala 2.11-2.13 and Akka 2.6 2.12-2.13 while the lagom support modules are more selective in what they work with?

marcospereira:
Hey @johanandren,

> Could be the time we need to split the two so that the journal can stay compatible with Akka 2.5 + Scala 2.11-2.13 and Akka 2.6 2.12-2.13 while the lagom support modules are more selective in what they work with?

I was thinking about this yesterday and, while I agree that having compatibility with Akka 2.5 and 2.6 is a good thing, it is pushing our users (Play & Lagom users at least) in some non-ideal situations. For example, Play and Lagom force the use of Akka 2.6, but some modules that depend on both Play/Lagom **and** Akka are causing [mixed versions](https://doc.akka.io/docs/akka/current/common/binary-compatibility-rules.html#mixed-versioning-is-not-allowed) since Akka modules depend on 2.5 instead.

And now we have this mismatch between Scala versions as well.

Does it make sense for Akka modules to cut new major/minor versions depending on 2.6 and dropping Scala 2.11 so that we can avoid these mismatches for Play and Lagom users?

lagom-persistence-couchbase/scaladsl/src/test/scala/com/lightbend/lagom/scaladsl/persistence/couchbase/CouchbaseReadSideSpec.scala
ignasi35: gracefully shutting down is delegated to Coordinated Shutdown, so maybe this test needs to run Coordinated Shutdown (even though a plain `actorSystem.terminate` will do too)

johanandren:
Given that this was released as a 1.0 during the Akka 2.5 lifecycle and that 2.6 is not yet released I don't think we can drop 2.6 support. I think dropping Scala 2.11 would be ok though.

I'd like to keep this library minimal effort since it is so new and we don't know of a large user base yet, keeping parallell branches like we do for Cassandra seems like a waste of effort.

Releasing the Lagom modules here on a different cadence with a different supported set of Scala + Akka seems like the right thing to do, not sure how big effort that would be - would we need a separate repo for example (we already sort of have done that with the connector living in Alpakka nowadays).

I think the transitive version-downgrade is a separate (an bigger) issue that we may have to discuss across teams.

marcospereira:
This is outdated now.

PR #223, opened 2019-08-09
Update to the new Lightbend Platform docs location

TimMoore:


TimMoore:
The Travis CI failure looks like another case of the flaky network issues we've seen lately. I don't have permission to re-run the job myself.

PR #227, opened 2019-10-14
Scala 2.12.10 and 2.13.1

ennru:

PR #234, opened 2020-01-17
Scala versions

ignasi35:
Supersedes #227

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,7 +58,7 @@ jobs:
     - env:
       - JDK="adopt@~1.11-0"
       - PRE_CMD="docker-compose -f docker/couchbase-1-node/docker-compose.yml up -d couchbase"
-      - CMD=";++2.12.8 test" # note that JDK 11 + Java requires Scala 2.12 because of extension getter ambiguity
+      - CMD=";++2.12.10 test" # note that JDK 11 + Java requires Scala 2.12 because of extension getter ambiguity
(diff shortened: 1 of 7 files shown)


johanandren:
Build fail is formatting of sbt files

ignasi35:
Ok, seems we're finally getting to the meat of it:

```
[error] /home/travis/build/akka/akka-persistence-couchbase/core/src/main/scala/akka/persistence/couchbase/internal/CouchbaseSchema.scala:327:60: object JavaConverters in package collection is deprecated (since 2.13.0): Use `scala.jdk.CollectionConverters` instead
```

.travis.yml
```diff
@@ -58,7 +58,7 @@ jobs:
     - env:
       - JDK="adopt@~1.11-0"
       - PRE_CMD="docker-compose -f docker/couchbase-1-node/docker-compose.yml up -d couchbase"
-      - CMD=";++2.12.8 test" # note that JDK 11 + Java requires Scala 2.12 because of extension getter ambiguity
+      - CMD=";++2.12.10 test" # note that JDK 11 + Java requires Scala 2.12 because of extension getter ambiguity
```
ennru: This must be in sync with sbt: 2.12.11

ignasi35: Good catch!

ignasi35:
```suggestion
      - CMD=";++2.12.11 test" # note that JDK 11 + Java requires Scala 2.12 because of extension getter ambiguity
```

